Issue #24, opened 2014-02-21
Allow to define a default background color as part of a color map


sschuberth:
It would be nice if one could define a (named) default background color as part of a color map, e.g. "Black". In fact, one would probably need to be able to distinguish between the "dark" and "bright" flavors of a color. So as a first step, the according rows in the color map should be labeled "dark" / "bright" (or "low" / "high" or something), and then the background color could be chosen from a combo-box that contains "Black (dark)", "Black (bright)", "Red (dark)", "Red (bright)" and so on.

ejelly:
Finally got to it. #41 contains a solution. It's per job as opposed to per colormap, but is there any convincing reason to have it per colormap?

sschuberth:
IIRC my motivation to request this to be per colormap was consistency. It seemed to make more sense to me to define all colors in the same place, and in a way that applies to all jobs that use a specific colormap. For example, I for sure would like all of my jobs to use black as the default background color and white as the default foregound color. It would be tedious if I would need to go through all job configs for this instead of just changing the colormap that of all of the jobs use.

sschuberth:
Also, I just read about issue #39. I realize that issue is *not* about adding another color preset, but if it was, and if default bg and fg color were defined as part of the preset, one could fake the visual appearance of a Powershell by setting the default bg color to #012456 and the default fg color to #EEEDF0 as part of the Powershell preset.

ejelly:
@dblock, do you agree? I can easily change it to be a part of the colormap instead.

dblock:
I think I do.

ejelly:
In the end I think you're right, it does make more sense as part of the colormap.

ejelly:
@sschuberth, #41 was just merged into master. Can you try it out?

sschuberth:
I've compiled the plugin and installed it into a public Jenkins instance. Unfortunately, it does not seem to fully work, see https://dscho.cloudapp.net/job/sdk-create-packages/67/console. I've used the xterm colormap and changed the default bg / fg colors to black / white. This is what I see:

- The black bg color does only seem to apply to a few lines in the beginning.
- Black is not actually black, but dark gray.

ejelly:
Not reproducible in my setup, I see the black (or rather dark gray) background throughout the whole console output, and it also seems to be fully enclosed by the div. Can you tell us what browser you are using and attach a screenshot?
[...]

PR #41, opened 2015-01-25
Default colors

ejelly:
To close issue #24, I implemented the ability to configure a default background and foreground color. If at least one of those is set, the console output will be contained within a div with the appropriate style attribute set. Note that #24 called for setting this in the color map, but I opted for a per-job configuration instead.

### DIFF
--- a/src/main/java/hudson/plugins/ansicolor/AnsiColorBuildWrapper.java
+++ b/src/main/java/hudson/plugins/ansicolor/AnsiColorBuildWrapper.java
@@ -51,6 +51,7 @@ import org.kohsuke.stapler.StaplerRequest;
  * 
  * @author Daniel Doubrovkine
  */
+@SuppressWarnings("unused")
 public final class AnsiColorBuildWrapper extends BuildWrapper {
 
 	private final String colorMapName;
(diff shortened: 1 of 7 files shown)


ejelly:
By the way, I did not update the README since I think the configuration is self-explanatory.

src/main/java/hudson/plugins/ansicolor/AnsiColorBuildWrapper.java
```diff
@@ -86,31 +99,35 @@ public Environment setUp(AbstractBuild build, Launcher launcher,
...
-            }
-
-            @Override
-            public void close() throws IOException {
-                ansi.close();
-                logger.close();
-                super.close();
-            }
-        };
+		if (logger != null)
```
dblock: Make this a bit more readable with some `{}` or maybe even with `if (logger == null) { return ... }`

src/main/java/hudson/plugins/ansicolor/AnsiColorBuildWrapper.java
```diff
@@ -86,31 +99,35 @@ public Environment setUp(AbstractBuild build, Launcher launcher,
...
+				}
+
+				@Override
+				public void close() throws IOException {
+					ansi.close();
+					logger.close();
+					super.close();
+				}
+			};
+		else
```
dblock:
```
else {
  return null;
}
```

dblock:
Looks good, can you fix the cosmetics and squash your commits please? I'll merge.

ejelly:
I changed the default fg/bg attributes to be part of the colormap, changed the "None" text do "Jenkins Default" (to make it clearer that you don't have to select something if you want ansicolor to leave the default foreground and background alone) and squashed the commits.

dblock:
Very nice, merging.

dblock:
You might want to update README maybe with a newer configuration screenshot, maybe with some notes on fg/bg colors?

ejelly:
Okay, I'll do that.

PR #42, opened 2015-01-27
Miscellaneous fixes and improvements

sschuberth:
These occurred to me when discussing PR #41.

### DIFF
--- a/src/main/java/hudson/plugins/ansicolor/AnsiColorMap.java
+++ b/src/main/java/hudson/plugins/ansicolor/AnsiColorMap.java
@@ -47,14 +47,14 @@ public final class AnsiColorMap implements Serializable
     @DataBoundConstructor
     public AnsiColorMap(
         String name,
-        String black, String red, String green, String blue, String yellow, String magenta, String cyan, String white,
-        String blackB, String redB, String greenB, String blueB, String yellowB, String magentaB, String cyanB, String whiteB,
+        String black, String red, String green, String yellow, String blue, String magenta, String cyan, String white,
+        String blackB, String redB, String greenB, String yellowB, String blueB, String magentaB, String cyanB, String whiteB,
         Integer defaultForeground, Integer defaultBackground) {
 
         this(
             name,
-            colorArray(black,red,green,blue,yellow,magenta,cyan,white),
-            colorArray(blackB,redB,greenB,blueB,yellowB,magentaB,cyanB,whiteB),
+            colorArray(black, red, green, yellow, blue, magenta, cyan, white),
+            colorArray(blackB, redB, greenB, yellowB, blueB, magentaB, cyanB, whiteB),
             defaultForeground, defaultBackground);
     }
 
@@ -76,8 +76,8 @@ public final class AnsiColorMap implements Serializable
     public String getBlack() { return normalMap[0]; }
     public String getRed() { return normalMap[1]; }
     public String getGreen() { return normalMap[2]; }
-    public String getBlue() { return normalMap[3]; }
-    public String getYellow() { return normalMap[4]; }
+    public String getYellow() { return normalMap[3]; }
+    public String getBlue() { return normalMap[4]; }
     public String getMagenta() { return normalMap[5]; }
     public String getCyan() { return normalMap[6]; }
     public String getWhite() { return normalMap[7]; }
@@ -85,8 +85,8 @@ public final class AnsiColorMap implements Serializable
     public String getBlackB() { return brightMap[0]; }
     public String getRedB() { return brightMap[1]; }
     public String getGreenB() { return brightMap[2]; }
-    public String getBlueB() { return brightMap[3]; }
-    public String getYellowB() { return brightMap[4]; }
+    public String getYellowB() { return brightMap[3]; }
+    public String getBlueB() { return brightMap[4]; }
     public String getMagentaB() { return brightMap[5]; }
     public String getCyanB() { return brightMap[6]; }
     public String getWhiteB() { return brightMap[7]; }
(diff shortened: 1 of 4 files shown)


dblock:
Thank you.

Can you please fix the tabs/spaces difference? There's only a small set of things changed here, but the diff is huge. Compare the diff with [this](https://github.com/dblock/jenkins-ansicolor-plugin/pull/42/files?w=1).

dblock:
I see that first commit (spaces -> tabs). I would be ok with that actually, @ejelly?

ejelly:
Yes, absolutely, it's good to have consistency in whitespaces.

Generally the pull requests looks fine to me.

Only issue: It is incorrect to let the color map directly return the color string (i.e. the actual hex color triplet) for the default colors, especially the default fg color, instead of the index. The reason is that ANSI codes actually cannot directly address the full 4 bit foreground low+high intensity color map, like VGA attributes can. Instead, the BOLD escape command switches between low and high intensity color maps, and the default foreground/background only specify the colors (resp. their indices), not the default intensity.

Right now this problem is not exposed because there is actually no way to access the bright color map at all, it sits there useless!

Now that I think of it, the choice whether the BOLD code switches to an actual bold typeface or to the high intensity color map instead could be another flag of the color map. VGA does not have bold typefaces and switches to high intensity, while xterm (as far as I remember) switches to bold.

But that is a separate new feature, and I'm fine with this being merged, modulo the change that returns the default colors als hex triplet string instead of as color index of course.

What do you think? It's entirely possible that I've overlooked something above, terminals have always been messy 8)

sschuberth:
> There's only a small set of things changed here, but the diff is huge.

That's why you should never look at the combined diff of the PR, but at the diffs of the individual commits ;-)

> I see that first commit (spaces -> tabs).

Right. FYI, I'd also be fine with using spaces instead of tabs (which I personally prefer), it just should be consistent, and I chose tabs in this case as that seemed to be the dominant style.

sschuberth:
> It is incorrect to let the color map directly return the color string

Did you notice that I introduced `getDefaultFore/BackgroundIndex()` to return the index? In other words, I've renamed the existing `getDefaultFore/Background()` to `getDefaultFore/BackgroundIndex()` (since they indeed do not return the color but the index) and introduced new `getDefaultFore/Background()` methods that return the color directly.

I admit this change is not strictly necessary, it was more or less a left-over of some other refactoring I was planning to do but that I discarded for now.

So, should I keep it or remove it?

sschuberth:
> Yes, absolutely, it's good to have consistency in whitespaces.

Do you want me to extend that to all *.java files in the project? So far I've only reformatted those files that I touched for the other changes. And if so, again the question: all tabs, or all spaces?

dblock:
My humble 0.02c opinion on tabs/spaces is to make a single separate PR with just those changes all over the project.

sschuberth:
> My humble 0.02c opinion on tabs/spaces is to make a single separate PR with just those changes all over the project.

I'll happily do it all over the project. But what's the benefit of a separate PR vs. separate commit for that?

ejelly:
> Did you notice that I introduced getDefaultFore/BackgroundIndex() to return the index? In other words, I've renamed the existing getDefaultFore/Background() to getDefaultFore/BackgroundIndex() (since they indeed do not return the color but the index) and introduced new getDefaultFore/Background() methods that return the color directly.

I understand and I do like the other changes, but the reason I think the methods returning String shouldn't be there at all: The state of the terminal is fully contained in `AnsiHtmlOutputStream`, that's what knows what's going on. If we later actually do have means to switch the current foreground color to the high intensity color map, getDefaultForeground() would still return the low intensity color, despite the BOLD/INTENSITY attribute being set. Try it out:

    echo -e 'plain\e[1mbold\e[32mboldred\e[39mresetfg\e[0mresetall'

In my Terminal.app, the foreground reset code `\e[39m` resets the color to its default, but the BOLD attribute remains set! It will be `AnsiHtmlOutputStream` that ultimately decides what actual color to put out, as we want the `AnsiColorMap` to remain stateless.

> I admit this change is not strictly necessary, it was more or less a left-over of some other refactoring I was planning to do but that I discarded for now.
> So, should I keep it or remove it?

Given that it will introduce a subtle bug once the BOLD attribute switches the intensity, please remove it. I know it seems a bit tedious for such a little thing, but, again, terminal emulators are kludgy 8)

ejelly:
Tabs/spaces: I'm afraid I may have accidentally introduced the tabs that are there in my recent commits due to messed up IDE settings, sorry for that.

In any case I also prefer spaces, the reason being that they are fixed width and so won't mess up any alignments.

There is also the very neat "tabs for indentation, spaces for alignment" which permits the use of variable width tabs and still have any alignment matching, but switching over to that seems like more trouble than it's worth.

> But what's the benefit of a separate PR vs. separate commit for that?

I guess it keeps your PR with your actual functional changes clean and the diff much easier to read if there aren't any intermingled whitespace changes (I know that you can tell git diff to ignore them, but I don't know about the github UI).

Thanks again!

sschuberth:
> Given that it will introduce a subtle bug once the BOLD attribute switches the intensity, please remove it.

Done. I'll push a new iteration of this PR once my other PR for the whitespace stuff is merged.

> I guess it keeps your PR with your actual functional changes clean and the diff much easier to read

That's why I was suggesting to not read the combined diff of all commits, but the diffs for each commit individuall when reviewing the PR ;-) Anyway, here's the PR with only the whitespace changes: https://github.com/dblock/jenkins-ansicolor-plugin/pull/45.

sschuberth:
So here's my updated version rebased against the current master and with the commit dropped that introduces `getDefaultFore/BackgroundIndex()`.

ejelly:
Thanks, merged!

PR #48, opened 2015-02-03
suspend tags at end of every output chunk to work around IE workaround

ejelly:
This closes and reopens the elements for every output chunk, effectively getting rid of the problem incurred by Jenkins' IE workaround.


sschuberth:
I can confirm this fixes the background color for real-time console updates for me, too. Thanks!

dblock:
This creates a span inside a span. Timestamper is fairly common, I think you should fix this before merging.

dblock:
@ssbarnea See comments above. Someone needs to finish this PR.

PR #54, opened 2015-07-31
Implementation of a SimpleBuildWrapper for use in workflows. Change i…

qvicksilver:
…n parent version due to "Build wrappers" headline here: https://github.com/jenkinsci/workflow-plugin/blob/cd9b86e225c772084c39ebb110275bf076f010ad/COMPATIBILITY.md


dblock:
This is great. I would merge with tests, please!

qvicksilver:
I read about that but noticed there was no test for you build wrapper so I thought I might get away with it ;) But I'll produce some test and update the PR.

dblock:
Well now you also have to write a test for the build wrapper :)

omehegan:
Can we get this merged? I am itching to have this support as I convert projects to Workflow. I love AnsiColor!

pom.xml
```diff
@@ -89,6 +89,11 @@
       <artifactId>jansi</artifactId>
       <version>1.9</version>
     </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci.plugins.workflow</groupId>
+      <artifactId>workflow-aggregator</artifactId>
+      <version>1.8</version>
```
jglick: This should be a `test`-scoped dependency. (Or deleted, if you do not plan to add an actual `JenkinsRule`-based test verifying Workflow compatibility.)

src/main/java/hudson/plugins/ansicolor/AnsiColorSimpleBuildWrapper.java
```diff
@@ -0,0 +1,158 @@
...
+import java.io.Serializable;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import jenkins.tasks.SimpleBuildWrapper;
+import org.apache.commons.lang.builder.HashCodeBuilder;
+import org.kohsuke.stapler.DataBoundConstructor;
+
+public class AnsiColorSimpleBuildWrapper extends SimpleBuildWrapper implements Serializable {
```
jglick: It suffices to just change the superclass of the existing `AnsiColorBuildWrapper`. (A `SimpleBuildWrapper` can be used by traditional projects too.)

src/main/java/hudson/plugins/ansicolor/AnsiColorSimpleBuildWrapper.java
```diff
@@ -0,0 +1,158 @@
...
+    
+    private static final long serialVersionUID = 1L;
```
jglick: Why is this being made `Serializable`? It should never be serialized AFAIK.

dblock:
Sorry, I didn't get a notification from Github from the update - thanks for writing the tests. I think some questions from @jglick seem legit, care to take a look please @qvicksilver. If you don't mind squashing commits too.

pom.xml
```diff
@@ -5,7 +5,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>1.424</version>
+    <version>1.609.1</version>
```
dblock: Is this now requiring a much newer version of Jenkins? I don't think that's what we want unless there's a strong reason.

pom.xml
dblock: Thanks for spotting this.

src/main/java/hudson/plugins/ansicolor/AnsiColorSimpleBuildWrapper.java
dblock: Yep, would be a lot less copy-paste.

pom.xml
qvicksilver: Agreed the comments by @jglick do make sense and I'll do something about them as soon as I get the time.

The "Plugin Compatibility with Workflow"-documentation (https://github.com/jenkinsci/workflow-plugin/blob/master/COMPATIBILITY.md#build-wrappers-1) states that "To add support for a BuildWrapper, depend on Jenkins 1.599+ (typically 1.609.1), and implement SimpleBuildWrapper, following the guidelines in its Javadoc." so thats the reason for depending on 1.609.1. I guess it is a fairly strong reason.

dblock: :+1:

pom.xml
qvicksilver: Changed scope and implemented a JenkinsRule based test in last commit.

dblock:
Merging.

jglick:
Retroactively filed as [JENKINS-30528](https://issues.jenkins-ci.org/browse/JENKINS-30528).

PR #55, opened 2015-09-23
Fixed up documentation regarding workflow support

qvicksilver:
Forgot to update the documentation regarding workflows with the changes from yesterday, here is the fixup

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,12 +81,12 @@ RSpec.configure do |config|
 end
 ```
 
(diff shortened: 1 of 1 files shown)


dblock:
If this works well we should do a release?

qvicksilver:
That might be a good idea, it's been a while since the last release and I think workflows are a very nice new Jenkins feature worth supporting.

PR #56, opened 2015-09-23
set `--tty` per argument

MartinNowak:


dblock:
So that Ruby part was not necessary? Isn't it something that you _could_ have instead of `.rspec`?

MartinNowak:
Right the ruby part did the same as passing --tty and yes you can put `--tty --colour` in an .rspec file instead of passing it as arguments.

PR #59, opened 2015-10-16
AnsiHtmlOutputStreamTest: Fix testJapanese() on Windows

sschuberth:
Do not rely on the default encoding to be UTF-8.


dblock:
Makes sense.

PR #58, opened 2015-10-16
AnsiColorMap: Change the serialVersionUID

sschuberth:
This is a fixup to commit 4ef902b that renamed member variables. The new
serialVersionUID was calculated using Java's "serialver" tool.

This is supposed to fix #57.

PR #60, opened 2015-10-17
Make Travis CI test on Mac OS X in addition to Linux

sschuberth:
This is to test what @omehegan said [here](https://github.com/dblock/jenkins-ansicolor-plugin/pull/58#issuecomment-148846633).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,2 +1,7 @@
 language: java
 
+matrix:
+  include:
+    - os: linux
+    - os: osx
+      env: JAVA_HOME=$(/usr/libexec/java_home)
--- a/src/test/java/hudson/plugins/ansicolor/AnsiColorBuildWrapperTest.java
+++ b/src/test/java/hudson/plugins/ansicolor/AnsiColorBuildWrapperTest.java
@@ -55,7 +55,7 @@ public class AnsiColorBuildWrapperTest {
                         "node {\n"
                         + "  wrap([$class: 'AnsiColorBuildWrapper', 'colorMapName': 'XTerm', 'defaultFg': 1, 'defaultBg': 2]) {\n"
                         + "    sh(\"\"\"#!/bin/bash\n"
-                        + "      echo -e '\\\\e[31mred\\\\e[0m'\"\"\"\n"
+                        + "      printf 'The following word is supposed to be \\\\e[31mred\\\\e[0m\\\\n'\"\"\"\n"
                         + "    )\n"
                         + "  }\n"
                         + "}"
--- a/src/test/java/hudson/plugins/ansicolor/AnsiColorBuildWrapperTest.java
+++ b/src/test/java/hudson/plugins/ansicolor/AnsiColorBuildWrapperTest.java
@@ -63,7 +63,8 @@ public class AnsiColorBuildWrapperTest {
                 story.j.assertBuildStatusSuccess(p.scheduleBuild2(0));
                 StringWriter writer = new StringWriter();
                 p.getLastBuild().getLogText().writeHtmlTo(0L, writer);
-                assertTrue(writer.toString().matches("(?s).*<span style=\"color: #CD0000;\">red</span>.*"));                        
+                String html = writer.toString();
+                assertTrue("Failed to match color attribute in following HTML log output:\n" + html, html.matches("(?s).*<span style=\"color: #CD0000;\">red</span>.*"));
             }
         });
     }


dblock:
Any luck with the failing test?

sschuberth:
No luck yet fixing the test suite on Mac OS X. Over [here](https://github.com/dblock/jenkins-ansicolor-plugin/pull/58#issuecomment-148949264) @qvicksilver was mentioning that something might be wrong with the test itself.

dblock:
I played with this for a bit. I get a clearer error:

```
java.lang.UnsupportedOperationException: no known implementation of class jenkins.tasks.SimpleBuildWrapper is named AnsiColorBuildWrapper
	at org.jenkinsci.plugins.workflow.structs.DescribableHelper.coerce(DescribableHelper.java:233)
	at org.jenkinsci.plugins.workflow.structs.DescribableHelper.buildArguments(DescribableHelper.java:186)
	at org.jenkinsci.plugins.workflow.structs.DescribableHelper.instantiate(DescribableHelper.java:91)
	at org.jenkinsci.plugins.workflow.steps.StepDescriptor.newInstance(StepDescriptor.java:103)
	at org.jenkinsci.plugins.workflow.cps.DSL.invokeMethod(DSL.java:133)
	at org.jenkinsci.plugins.workflow.cps.CpsScript.invokeMethod(CpsScript.java:98)
	at org.codehaus.groovy.runtime.callsite.PogoMetaClassSite.call(PogoMetaClassSite.java:45)
	at org.codehaus.groovy.runtime.callsite.CallSiteArray.defaultCall(CallSiteArray.java:42)
	at org.codehaus.groovy.runtime.callsite.AbstractCallSite.call(AbstractCallSite.java:108)
	at com.cloudbees.groovy.cps.sandbox.DefaultInvoker.methodCall(DefaultInvoker.java:15)
	at WorkflowScript.run(WorkflowScript:2)
	at Unknown.Unknown(Unknown)
```

It can't find `AnsiColorBuildWrapper`? Why would that work on Linux?

dblock:
I opened https://issues.jenkins-ci.org/browse/JENKINS-31029, @qvicksilver if you have a solution, speak up!

qvicksilver:
@dblock sorry, should have reported in. But no I don't have a solution and can't find anything wrong with the test either.

hammertoe:
I was having same trouble with AnsiColorBuildWrapper not being found on OSX build slave. But changing declaration from:

`wrap([$class: 'AnsiColorBuildWrapper', 'colorMapName': 'xterm']) {`

to:

`wrap([$class: 'AnsiColorBuildWrapper', 'colorMapName': 'XTerm', 'defaultFg': 1, 'defaultBg': 2]) {`

Fixed it for me.

sschuberth:
@hammertoe Where exactly did you change that code? I was unable to find it. Also, is it sufficient to only change `xterm` to `XTerm`, i.e. is it about case-sensitivity?

hammertoe:
I changed it in my Jenkinsfile and yes, it looks like just changing xtem to XTerm appears to be sufficient. ie. it does indeed seem to be about case sensitivity.

-Matt

dblock:
I'm looking forward to a green build in this PR, with commits squashed ;)

sschuberth:
I'm still a bit unclear what to fix and where. If the issue is about case-sensitivity, and considering what's different between Linux and Mac, one might think it's some kind of file system issue, as most Linux file systems are case-sensitive while on Mac it's by default case-insensitive. But then why would it succeed on (strict) Linux and fail on (lenient) Mac? Also, the test passes fine for me on Windows, which also has a case-insensitive file system by default.

Finally, the test is already using [XTerm](https://github.com/dblock/jenkins-ansicolor-plugin/blob/master/src/test/java/hudson/plugins/ansicolor/AnsiColorBuildWrapperTest.java#L56), which according to @hammertoe should fix the issue. So what am I missing here?

sschuberth:
@dblock How were you able to get this message you've mentioned above:

> java.lang.UnsupportedOperationException: no known implementation of class jenkins.tasks.SimpleBuildWrapper is named AnsiColorBuildWrapper

I've Googled for it and found this other [similar issue](http://comments.gmane.org/gmane.comp.java.jenkins.user/35801), but without any answers.

sschuberth:
[I'm one step further](https://github.com/dblock/jenkins-ansicolor-plugin/pull/60/commits/2ee5b213d9498be30ed4d23e131d9a02b090cada): On Mac OS X, `echo` does not support, `-e`. But now I'm getting

    WARNING: Failed to add HTML markup '...'
    java.io.IOException: Stream Closed

sschuberth:
I've fixed it for good! :-) @hammertoe's issue was not related. Maybe writing "xterm" in lower case is another way of triggering this issue, but the Mac-specific cause was the use of `echo` with `-e`. When using `printf`, a final `\n` needs to be added (which `echo` does automatically) to avoid the "Stream Closed" IOException.

dblock:
:+1: merging this, thanks for debugging it @sschuberth

PR #73, opened 2016-06-14
Promote pipeline configuration in README

abrom:
Documentation for configuring pipeline (workflow) builds was given 2nd class citizen status and as such it isn't entirely clear from first glance it worked

Also updated to refer to newer pipeline plugin name (as well as older workflow name)


dblock:
Fair enough.

Issue #89, opened 2017-03-03
Italic attribute and some others not recognized


Joe-Merten:
I had created a test matrix which is located within the comments of [this python script](https://github.com/Joe-Merten/Jfe/blob/master/Framework/tools/AnsiColors.py).

Most important lacks for my opinion:

- `esc[3m` / `esc[23m`  italic on / off
- `esc[7m` / `esc[27m`  inverse on / off
  - see also: https://github.com/dblock/jenkins-ansicolor-plugin/issues/64
- `esc[9m` / `esc[29m`  strikeout on / off
- `esc[90m` … `esc[97m` and `esc[100m` … `esc[107m`  bright colors
  - see also: https://github.com/dblock/jenkins-ansicolor-plugin/issues/16
- `esc[38;…m` / `esc[48;…m`  256 colors and 24 bit colors
  - see also: https://github.com/dblock/jenkins-ansicolor-plugin/issues/44
  - and: https://issues.jenkins-ci.org/browse/JENKINS-24378
- `esc[28m`  reveal / conceal off

How to reproduce: Just put the above mentioned [python script](https://github.com/Joe-Merten/Jfe/blob/master/Framework/tools/AnsiColors.py) into a Jenkins freestyle or pipeline job with AnsiColor plugin enabled and then check the console output of this job in the web browser.

dblock:
The next step is to turn this into a test in this project, PRs welcome.

Joe-Merten:
I might contribute some test and / or code. What's a good point to start?
I have some knowledge in Java, C++ and some other languages, but have not doing plugin development before.

Joe-Merten:
Hmm, saw https://github.com/dblock/jenkins-ansicolor-plugin/pull/72 which is listed under 0.4.3 in the changelog. But for me, `esc[91m` etc. still won't work.

Joe-Merten:
With PR #90, this is done for:

- `esc[3m` / `esc[23m`  italic on / off
- `esc[9m` / `esc[29m`  strikeout on / off
- `esc[28m`  reveal / conceal off

and additional:

- `esc[51m` / `esc[54m`  framed on / off
- `esc[53m` / `esc[55m`  overlined on / off

Still todo:

- `esc[7m` / `esc[27m`  inverse on / off
  - see also: https://github.com/dblock/jenkins-ansicolor-plugin/issues/64
- `esc[90m` … `esc[97m` and `esc[100m` … `esc[107m`  bright colors
[...]

Issue #64, opened 2015-12-02
Inverted color not being shown


lucasvc:
As [read here](http://misc.flogisoft.com/bash/tip_colors_and_formatting), the escape sequence `\e[7m` allow to invert the colors. In example:

``` bash
RED='\e[31m'
INVERTED='\e[7m'
NORMAL='\e[0m'

if [ ! -f $HOME/_initilized ]; then
    echo -e "${RED}Some message\nMarking ${INVERTED}this text${NORMAL}."
    exit 1
fi
```

In bash `this text` is shown in red background and 'nocolor' foreground.
But the jenkins output is all in red color.

Joe-Merten:
I think this could be closed due to #94. But maybe someone do some more testing before closing, e.g. even with different schemes etc.

JoeMerten:
Fixed in 0.5.0

PR #94, opened 2017-03-07
Added support for negative / inverse text

Joe-Merten:
See also #89, #64.


Joe-Merten:
Yep, much changes. Main reason is to correctly handle fg / bg color changes inside of `esc[7m` … `esc[27m`.

Regarding the question about how to obtain the default colors, for the fg color (which became bg color upon `esc[7m` I decide for currentColor trick mentioned at [SO](http://stackoverflow.com/a/42586457/2880699). Well, this makes the implementation more complex, but gives the best results.

dblock:
👍  

also YOLO

Issue #95, opened 2017-03-08
High Indensity Colors won't work because of Jansi version conflict


Joe-Merten:
See also:

- #16
- #91
- http://stackoverflow.com/questions/42648284/how-to-cleanly-solve-jansi-jar-version-conflict-in-jenkins-plugin
- #72
- https://github.com/fusesource/jansi/pull/72

### Tested on:

- both Kubuntu 16.04 & 14.04 hosts
- using a simple freestyle project, executing the shell command: `printf "\e[31mRED\e[92mLIME\e[m\n"`

### Jenkins 1.653

Starting with a fresh new installation of Jenkins 1.653 and AnsiColor 0.4.3, I see that the `\e[31m` has been correctly processed, but `\e[92m` not:

- ![fail](https://cloud.githubusercontent.com/assets/6552470/23713313/cfff7832-0425-11e7-8539-e908012589ba.png)

Checking jenkins home directory for conflicting jansi versions:

    $ find . -name "*jansi*"
    ./.jenkins/plugins/ansicolor/WEB-INF/lib/jansi-1.12.jar
    ./.jenkins/war/WEB-INF/lib/jansi-1.9.jar

removing that disturbing 1.9:

    $ rm ./.jenkins/war/WEB-INF/lib/jansi-1.9.jar

and restart Jenkins. Now the `\e[92m` will also processed:
- ![ok](https://cloud.githubusercontent.com/assets/6552470/23713319/d68d1ede-0425-11e7-9b1f-a3bfd9cb272f.png)

→ That the facts, I'd already described in previous posts.

### Jenkins 2.49

Now, I tried the same with a most recent Jenkins 2.49, even completely fresh installed.
After installation and setup of above mentioned tiny buildjob, I got the same result as with 1.653: The `\e[31m` has been correctly processed, but `\e[92m` not.

Checking jenkins home directory for conflicting jansi versions:

    $ find . -name "*jansi*"
    ./.jenkins/plugins/ansicolor/WEB-INF/lib/jansi-1.12.jar
    ./.jenkins/war/WEB-INF/lib/jansi-1.11.jar

but removing that obvouisly disturbing 1.11 won't fix this issue here. And I even have not found a way to got it work with Jenkins 2.49.

### Possible solution

[...]

PR #72, opened 2016-05-30
Support high intensity ANSI colors #16 with jansi-1.12

marlene01:
pom.xml changes: - increased dependency of jansi to 1.12
                 - fixed the pom error with animal-sniffer plugin
AnsiHtmlOutputStream changes:
 - added method processSetForegroundColor with additional parameter boolean bright to handle the 90 to 97 colors.

Issue #44, opened 2015-01-27
Implement full xterm-256color range (probably needs JANSI changes)


ejelly:
This includes support for xterm's 256 color cube.

PierreR:
That would make `92` to be display as green ?
```
printf "\e[92mHello World\e[0m\n"
```
That would be awesome because I am missing this color badly ;-)

Joe-Merten:
Should be done now due to #96 / #98.

dblock:
Closing.

PR #96, opened 2017-03-08
Added support for xterm 256 colors and 24 bit colors

Joe-Merten:
This PR adds support for `esc[38;…m` / `esc[48;…m` for 256 colors and 24 bit foreground / background colors. See also:
- #89
- #44
- https://issues.jenkins-ci.org/browse/JENKINS-24378

Please note, that this needs [jansi PR #72](https://github.com/fusesource/jansi/pull/72) to build & work. I'd increased the jansi version dependency in the `pom.xml` to `1.14-SNAPSHOT` to build locally.

Further, we run into a version bump as described in #95 which leads also to make 256 colors and 24 bit colors don't work (even if we have successfully build and all plugin tests passed).

As I found a way to got it work with Jenkins 1.653 (by just deleting jansi-1.9.jar), I'd found no way to make it work with Jenkins 2.x. As dblock commented in #95, we might have to wait until even the Jenkins guys will pick up the latest jansi changes.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -88,7 +88,7 @@
     <dependency>
       <groupId>org.fusesource.jansi</groupId>
       <artifactId>jansi</artifactId>
-      <version>1.12</version>
+      <version>1.14-SNAPSHOT</version>
     </dependency>
     <dependency>
      <groupId>org.jenkins-ci.plugins.workflow</groupId>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 * [#90](https://github.com/dblock/jenkins-ansicolor-plugin/pull/90): Added missing handling for ATTRIBUTE_CONCEAL_OFF - [@Joe-Merten](https://github.com/Joe-Merten).
 * [#90](https://github.com/dblock/jenkins-ansicolor-plugin/pull/90): Added support for italic, strikeout, framed and overline attributes - [@Joe-Merten](https://github.com/Joe-Merten).
 * [#92](https://github.com/dblock/jenkins-ansicolor-plugin/pull/92): Fixed high intensity colors for both foreground and background - [@Joe-Merten](https://github.com/Joe-Merten).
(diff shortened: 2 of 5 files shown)


pom.xml
```diff
@@ -88,7 +88,7 @@
     <dependency>
       <groupId>org.fusesource.jansi</groupId>
       <artifactId>jansi</artifactId>
-      <version>1.12</version>
+      <version>1.14-SNAPSHOT</version>
```
dblock: I don't think we want to ship with an unshipped version of 1.14. I am not saying we shouldn't merge this, but we should be careful releasing with this.

dblock:
... and this doesn't pass build since the SNAPSHOT version is a private one.

Joe-Merten:
Ok, then I just have to wait for:
- [jansi #72](https://github.com/fusesource/jansi/pull/72) became merged
- jansi 1.15 release

Both might take a indefinite period of time.

And after this, I'll:
- change the pom.xml in this PR.
- send a PR to [jenkinsci](https://github.com/jenkinsci/jenkins) to also change their jansi version in pom.xml

CU next time, Joe.

dblock:
Thanks Joe! 

Maybe release what's on master though? It fixes a number of issues.

Joe-Merten:
In #95, you commented:

> … but maybe including JANSI as a built submodule (in a different namespace) is OK?

Could you help me to understand the idea behind this approch?
Then I'll give it a try to make _MoreColors_ work without waiting for Jansi & Jenkins repos (and I could refactor the code as soon as Jansi & Jenkins have been caught up).

dblock:
I mean checking out the source of Jansi as a submodile, then bulk replacing the namespace by something internal during the build and using that code.

Joe-Merten:
I'm a bit aware about git submodule because of trouble in the past (but however, I'm not git expert at all).
What about PR #98? I'd tested it even in bleeding edge jenkins 2.50 and it works as expected.

dblock:
I like it enough. Party on.

PR #98, opened 2017-03-16
More colors no jansi

Joe-Merten:
This PR is an addition to #96, but also drops the depentency to jansi (because of massive version trouble).

I'd just grabbed jansi's AnsiOutputStream.java and put it into ansicolor source tree, so it has been moved from package `org.fusesource.jansi` to `hudson.plugins.ansicolor`.

With this change, ext color support works regardless which jansi version is used by jenkins or other plugins.


dblock:
I think this is fine, we can always update from source. Lets go for it.

PR #107, opened 2017-05-11
Removing startup banner

jglick:
No other plugin that I know of prints a banner at startup; it is a little obnoxious. You can always advertise in `src/main/resources/index.jelly`.

@reviewbybees


alexanderrtaylor:
🐝 
This version of the plugin works for me and I agree about the banner(from what I have seen)

dblock:
I suppose you just want Jenkins to be boring :(

Please update CHANGELOG, https://github.com/jenkinsci/ansicolor-plugin/blob/master/CHANGELOG.md, we'll think about it.

dblock:
Just to be nitpicky, we have debugged the plugin by asking people to look for the message before. It's a confirmation that it actually gets loaded.

jglick:
> we have debugged the plugin by asking people to look for the message before

Does it not suffice to check `/pluginManager/installed`? Or, in complex cases, ask them to install `support-core` and send you a bundle?

dblock:
Spoken like a true engineer @jglick. Also they can attach a remote debugger ;)

jglick:
The list of installed plugins is clearly visible from the UI.

dblock:
I still don't get it. If Jenkins doesn't want plugins to advertise themselves on startup why is this available at all? What's this feature hurting and who is it annoying?

dblock:
Btw, I am merging this cause I am sensitive to the argument that "no other plugin does that", but IMO I am surprised this is an issue to anyone.

weakcamel:
Personally, I agree with the decision to remove the banner.

http://www.linfo.org/rule_of_silence.html

Logging that a plugin has been loaded isn't useful or interesting, and definitely shouldn't be surprising :-)

dblock:
Since this thread is still alive - sometimes we want computers to just be more fun.

weakcamel:
@dblock Absolutely, and that's a good thing - see for instance easter eggs in almost all software.

IMHO, it's just not the most fortunate way to introduce the fun factor:

- you review logs when something's broken, and then you appreciate them to be as concise as possible
- logging on initialization can cause pain in general; just 2 days ago my friend told me the story of struggling to combine and filter out logs from several libraries into one sensible logger and his biggest pain was one library which logged some output on ... `import`, before he could set it all up.

Just my $0.02 and sorry for pulling zombies out of graves! :-)

PR #109, opened 2017-07-06
Workflow step env var

mkobit:
Fixes https://github.com/jenkinsci/ansicolor-plugin/issues/51 for pipeline users

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 
 * [#100](https://github.com/jenkinsci/ansicolor-plugin/pull/100): Migrated hosting to github.com/jenkinsci - [@JoeMerten](https://github.com/JoeMerten) & [@dblock](https://github.com/dblock).
 * [#101](https://github.com/jenkinsci/ansicolor-plugin/pull/101): Some exceptions during plugin install and following jenkins start - [@JoeMerten](https://github.com/JoeMerten).
+* [#109](https://github.com/jenkinsci/ansicolor-plugin/pull/109): Set `TERM` environment variable inside of the `ansiColor` step when using Jenkins pipelines
 * Your contribution here.
 
 0.5.0  (03/18/2017)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 
 * [#100](https://github.com/jenkinsci/ansicolor-plugin/pull/100): Migrated hosting to github.com/jenkinsci - [@JoeMerten](https://github.com/JoeMerten) & [@dblock](https://github.com/dblock).
 * [#101](https://github.com/jenkinsci/ansicolor-plugin/pull/101): Some exceptions during plugin install and following jenkins start - [@JoeMerten](https://github.com/JoeMerten).
-* [#109](https://github.com/jenkinsci/ansicolor-plugin/pull/109): Set `TERM` environment variable inside of the `ansiColor` step when using Jenkins pipelines
+* [#109](https://github.com/jenkinsci/ansicolor-plugin/pull/109): Set `TERM` environment variable inside of the `ansiColor` step when using Jenkins pipelines - [@mkobit](https://github.com/mkobit).
 * Your contribution here.
 
 0.5.0  (03/18/2017)
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>2.21</version>
+    <version>2.25</version>
   </parent>
 
   <repositories>
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>1.609.1</version>
+    <version>2.21</version>
   </parent>
 
   <repositories>
@@ -14,6 +14,10 @@
       <url>http://repo.jenkins-ci.org/public/</url>
     </repository>
   </repositories>
+
+  <properties>
+    <jenkins.version>1.642.3</jenkins.version>
+  </properties>
 
   <pluginRepositories>
     <pluginRepository>
@@ -88,13 +92,13 @@
     <dependency>
      <groupId>org.jenkins-ci.plugins.workflow</groupId>
      <artifactId>workflow-step-api</artifactId>
-     <version>1.7</version>
+     <version>2.12</version>
      <optional>true</optional>
     </dependency>
     <dependency>
       <groupId>org.jenkins-ci.plugins.workflow</groupId>
       <artifactId>workflow-aggregator</artifactId>
-      <version>1.8</version>
+      <version>2.5</version>
       <scope>test</scope>
     </dependency>
   </dependencies>
--- a/src/main/java/hudson/plugins/ansicolor/AnsiColorBuildWrapper.java
+++ b/src/main/java/hudson/plugins/ansicolor/AnsiColorBuildWrapper.java
@@ -176,7 +176,7 @@ public final class AnsiColorBuildWrapper extends SimpleBuildWrapper implements S
          */
         @Override
         public String getDisplayName() {
-            return Messages.plugin_DisplayName();
+            return Messages.DisplayName();
         }
 
         /**
--- a/src/main/java/hudson/plugins/ansicolor/AnsiColorBuildWrapper.java
+++ b/src/main/java/hudson/plugins/ansicolor/AnsiColorBuildWrapper.java
@@ -23,7 +23,6 @@
  */
 package hudson.plugins.ansicolor;
 
-import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import hudson.EnvVars;
 import hudson.Extension;
 import hudson.FilePath;
--- a/src/main/java/hudson/plugins/ansicolor/AnsiColorBuildWrapper.java
+++ b/src/main/java/hudson/plugins/ansicolor/AnsiColorBuildWrapper.java
@@ -23,6 +23,7 @@
  */
 package hudson.plugins.ansicolor;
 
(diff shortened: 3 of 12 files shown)


src/main/java/hudson/plugins/ansicolor/AnsiColorBuildWrapper.java
```diff
@@ -23,6 +23,7 @@
  */
 package hudson.plugins.ansicolor;
 
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
```
dblock: Is this necessary?

src/main/resources/hudson/plugins/ansicolor/Messages.properties
```diff
@@ -20,4 +20,4 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
 
-DisplayName = Color ANSI Console Output
+plugin_DisplayName=Color ANSI Console Output
```
dblock: Could you please explain this change?

pom.xml
```diff
@@ -5,7 +5,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>1.609.1</version>
+    <version>2.25</version>
```
dblock: This changes the minimum compatibility version of Jenkins if I am not mistaken, is this required?

pom.xml
```diff
@@ -15,6 +15,10 @@
     </repository>
   </repositories>
 
+  <properties>
+    <jenkins.version>1.642.3</jenkins.version>
```
dblock: Is this a new minimum version? Where is this used?

src/main/resources/hudson/plugins/ansicolor/Messages.properties
mkobit: I changed this because FindBugs was complaining about a method starting with a capital letter, so I prefixed it with the lowercase `plugin_`

I asked on the Jenkins IRC to see if there was a way to ignore generated because I am not familiar with the build process for Jenkins projects but didn't get an answer.

pom.xml
mkobit: Yes, I think this will bump it up because of how https://github.com/jenkinsci/plugin-pom brings in Jenkins versions.

pom.xml
mkobit: This would be the new minimum version. This property appears to be used in the parent pom for dependencies (the `org.jenkins-ci.plugins:plugin:2.25` version).

The updated dependency on `workflow-step-api` requires the Jenkins version as the base, and the tests fail with lots of classloading errors if at least this version isn't used.

There _may_ be some hackery that can done to avoid bringing in the new workflow API version for https://github.com/jenkinsci/ansicolor-plugin/pull/109/files#diff-8c1fae68c810376b2cd62d70313c4c3cR72 , but I am not confident in my Jenkins plugin development knowledge to make a decision on this stuff.

src/main/resources/hudson/plugins/ansicolor/Messages.properties
dblock: I am just not familiar with the whole findbugs business. Those are warnings though, right? I worry that this introduces some dependency that's not available at runtime or something like that?

pom.xml
dblock: I think that's OK, we can bump things, but we need to document the minimum new requirement clearly in CHANGELOG and README and probably increment a major version of the plugin in this PR.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 
 * [#100](https://github.com/jenkinsci/ansicolor-plugin/pull/100): Migrated hosting to github.com/jenkinsci - [@JoeMerten](https://github.com/JoeMerten) & [@dblock](https://github.com/dblock).
 * [#101](https://github.com/jenkinsci/ansicolor-plugin/pull/101): Some exceptions during plugin install and following jenkins start - [@JoeMerten](https://github.com/JoeMerten).
+* [#109](https://github.com/jenkinsci/ansicolor-plugin/pull/109): Set `TERM` environment variable inside of the `ansiColor` step when using Jenkins pipelines
```
dblock: Make this line look like others, it's missing your name.

src/main/java/hudson/plugins/ansicolor/AnsiHtmlOutputStream.java
```diff
@@ -193,6 +194,7 @@ private void closeTagsOfTypeFGBG() throws IOException {
         }
     }
 
+    @SuppressFBWarnings("SF_SWITCH_NO_DEFAULT")
```
dblock: This says there's a switch statement without a default case, maybe just fix that statement?

src/main/java/hudson/plugins/ansicolor/AnsiHtmlOutputStream.java
```diff
@@ -473,6 +476,7 @@ private String getRgbColor(int r, int g, int b) throws IOException {
         return "#" + String.format("%02X", r) + String.format("%02X", g) + String.format("%02X", b);
     }
 
+    @SuppressFBWarnings("UC_USELESS_CONDITION")
```
JoeMerten: I think that the useless condition is `if (paletteIndex < 256)`.
The SuppressFBWarnings annotation will suppress all useless condition warnings in this function, isn't it?
IMHO it should only suppress the single location of `if (paletteIndex < 256)`.

src/main/java/hudson/plugins/ansicolor/AnsiOutputStream.java
```diff
@@ -42,7 +42,7 @@
  */
 public class AnsiOutputStream extends FilterOutputStream {
 
-    public static final byte[] REST_CODE = resetCode();
+    private static final byte[] REST_CODE = resetCode();
```
JoeMerten: Just a hint: Please note, that `AnsiOutputStream.java` is copied from [jansi](https://github.com/fusesource/jansi/blob/master/jansi/src/main/java/org/fusesource/jansi/AnsiOutputStream.java) because of version conflict (see also #95 and #96).
Doing thus changes will make it harder to compare against the original in future.

dblock: Yeah, that was my feeling too, we should really leave this alone. But there's an actual "UTF-8" change in there too that seems to fix something? Should it go upstream?

JoeMerten: If someone has time to investigate how far this might be a real issue, it should be reported to jansi.

mkobit:
Sorry to let the linger when you guys were do fast with feedback! I will take another pass in the next couple days.

src/main/resources/hudson/plugins/ansicolor/Messages.properties
mkobit: My understanding is that it will not change anything at runtime. It is used at build time for checking for bugs - the annotation has `@Retention(RetentionPolicy.CLASS)` so I don't think there would be any way for runtime to have any different behavior. I think it is included by the newer parent plugin version which is why the build fails with.

If I run the `mvn verify` with the suppressions removed I get:

```
[INFO] ------------------------------------------------------------------------                                        
[INFO] BUILD FAILURE                                       
[INFO] ------------------------------------------------------------------------                                        
[INFO] Total time: 49.565 s                                
[INFO] Finished at: 2017-07-20T10:18:59-05:00              
[INFO] Final Memory: 63M/1070M                             
[INFO] ------------------------------------------------------------------------                                        
[ERROR] Failed to execute goal org.codehaus.mojo:findbugs-maven-plugin:3.0.4:check (findbugs) on project ansicolor: failed with 3 bugs and 0 errors  -> [Help 1]                                                                              
[ERROR]                                                    
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.                                    
[ERROR] Re-run Maven using the -X switch to enable full debug logging.                                                 
[ERROR]                                                    
[ERROR] For more information about the errors and possible solutions, please read the following articles:              
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoExecutionException
```

I see a way in the parent plugin POM to configure exclude rules at `${project.basedir}/src/findbugs/excludesFilter.xml` which may be an approach for a few of these, but I'm not sure.

mkobit: The exclude rules seems like it could be applied to classes. I found https://github.com/jenkinsci/branch-api-plugin/blob/master/src/findbugs/excludesFilter.xml which has:

```xml
<FindBugsFilter>
  <Match>
    <Class name="~.*\.Messages"/>
  </Match>
</FindBugsFilter>
```

We could also use that here?

src/main/java/hudson/plugins/ansicolor/AnsiHtmlOutputStream.java
mkobit: I think you are right on where this is happening - the `@SuppressFBWarnings` can't be applied to only that section unfortunately. 

I made a new commit that I'll push up shortly that hopefully clarifies this method and removes the suppression.

mkobit:
I pushed up a few changes to address a few FindBugs issues and tried to tackle `UC_USELESS_CONDITION`. I "reverted" the changes in `AnsiOutputStream` and excluded it from the FindBugs checks. I don't know if this was the best approach, but I thought it made more sense since it was mentioned that that class was copied from *jansi*.

I don't have a gerat idea for how to handle the `SF_SWITCH_NO_DEFAULT` - I could just add a `default` block that logs something but that doesn't seem entirely useful - any suggestions are very welcome!

dblock:
I vote for adding a default block instead of suppressing the warning. Raise an exception in there if you think the code is unreachable.

mkobit:
Sounds good - I'll take care of that today or tomorrow and then hopefully call this good to go!

dblock:
Merged.

dblock:
@mkobit Want to help out with the next release? https://github.com/jenkinsci/ansicolor-plugin/issues/103

PR #111, opened 2017-08-09
Filter out escape sequence 'character set' select

pmhahn:
Some terminals (xterm) supports two different 'character sets' ('G0' and
'G1'), which can be switched with 'SI' (shift in, '\017') and 'SO'
(shift out, '\016'). Each character set can be configured separately
and can be chosen from a list of pre-defined sets like 'ASCII Set' and
'Special Graphics'.

Running `TERM=xterm tput sgr0` to reset the terminal returns an escape
sequence starting with 'ESC ( 0', which selects the 'ASCII Set' for
'G0'. This currently is not understood by the Jenkins AnsiColor plugin.

Filter out those sequences, as they otherwise clutter the output.

<https://www.in-ulm.de/~mascheck/various/alternate_charset/> has a nice
description for characters sets

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,11 @@
 0.5.2 (Next)
 ============
 
+* [#111](https://github.com/jenkinsci/ansicolor-plugin/pull/111): Filter out escape sequence 'character set' select - [@pmhahn](https://github.com/pmhahn).
 * Your contribution here.
 
 0.5.1 (08/10/2017)
-============
+==================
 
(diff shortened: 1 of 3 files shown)


dblock:
Right on @JoeMerten, @pmhahn please do that and link it before we merge it here so it doesn't get lost in the shuffle.

daniel-beck:
Upstream issue: https://github.com/fusesource/jansi/issues/29

dblock:
For our PR here, this needs a CHANGELOG entry please @pmhahn. I can merge after.

pmhahn:
Updates the CHANGELOG.md again after you seem to have released already yesterday without pulling my change. Hope its okay now. Thanks.

dblock:
Merged, thanks.

PR #114, opened 2017-08-16
Ignore unknown SGR (Select Graphic Rendition) parameters

### LINKED ISSUE #112: Python Radish bdd tool crashes entire Jenkins job when coloring ansi output with AnsiColor 0.5.1
### Test environment

- Jenkins Version: 2.22
- AnsiColor Version: 0.5.1
- Hostsystem: Cent-OS

### Expected behavior

python radish bdd tool (A BDD testing tool for python, mimics Cucumber for Java -- installable with pip install radish-bdd) uses ansi coloring and it prints test scenarios colorfully to console

I'm unable to provide screenshots as they would contain proprietary information from my company's Jenkins build server.

We were trying to debug why our BDD test tool was crashing so badly without any warning all of a sudden (It worked yesterday). Turns out the Ansi color plugin was updated yesterday and introduced a regression. Turning off Ansi coloring in Jenkins re-enables the Job and it runs fine, but console output obviously looks terrible since there are un-interpreted ANSI symbols all over the place

### Actual behavior

It seems the upgrade from 0.5.0 to 0.5.1 creates some sort of regression when radish tries to print things to console in Jenkins. Note that Ansi coloring still works with cucumber (The Java bdd tool)

"Please provide small but complete code, e.g. shell script and / or Jenkinsfile"

As mentioned above, I am unable to provide any such code because it contains proprietary information :\ However, if you install the radish tool (https://github.com/radish-bdd/radish) and write a sample test and try to run it in the Test environment mentioned above, when you will run the radish command, the job will finish with an error, and there will be no console output past the command you just ran, not even a "Execute shell step marked build as failure"....It just fails. I was able to recreate the issue we were having in our builds which test our software following these steps

pmhahn:
like select primary/n-th alternate font

Fixes #112

Quick test:

    printf '\033[10m\n'

### DIFF
--- a/src/main/java/hudson/plugins/ansicolor/AnsiHtmlOutputStream.java
+++ b/src/main/java/hudson/plugins/ansicolor/AnsiHtmlOutputStream.java
@@ -357,7 +357,7 @@ public class AnsiHtmlOutputStream extends AnsiOutputStream {
     }
 
     // add attribute constants which are currently missing in jansi
-    // see also https://en.wikipedia.org/wiki/ANSI_escape_code
+    // see also <https://en.wikipedia.org/wiki/ANSI_escape_code#graphics>
     protected static final int ATTRIBUTE_STRIKEOUT       =  9;
     protected static final int ATTRIBUTE_ITALIC_OFF      = 23;
     protected static final int ATTRIBUTE_STRIKEOUT_OFF   = 29;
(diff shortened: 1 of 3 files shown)


dblock:
Thanks @pmhahn, merged.

dblock:
This also fixes https://github.com/jenkinsci/ansicolor-plugin/issues/113, right?

src/main/java/hudson/plugins/ansicolor/AnsiHtmlOutputStream.java
```diff
@@ -466,7 +469,7 @@ else switch (attribute) {
             closeTagOfType(AnsiAttrType.OVERLINE);
             break;
         default:
-            throw new IllegalStateException("Attribute " + attribute + " should not be reached");
+            break;
```
JoeMerten: Might comment why turned the exception into `break`?

pmhahn: See https://github.com/jenkinsci/ansicolor-plugin/issues/112#issuecomment-322742298 for the full analysis: https://github.com/jenkinsci/ansicolor-plugin/commit/2c09afa6729dc952a8fb99eeb9f23c15ae37cfdc changed the behavior to throw an IllegalStateException instead of ignoring unknown escape sequences. As only a fraction of existing escape sequences are handled, that is a very bad idea.

PR #119, opened 2017-10-09
add helper method to create AnsiHtmlOutputStream

jstrachan:
minor refactor to make it easier to create new AnsiHtmlOutputStream objects
to wrap an OutputStream for easier reuse of the ansicolor plugin in other jenkins plugins

e.g. I wanted to be able to wrap an OutputStream easily with an AnsiHtmlOutputStream inside the [updatebot plugin](https://github.com/jenkinsci/updatebot-plugin) but couldn't as the emitter stuff is not public.

So adding this reusable public static helper method makes it easier to reuse

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 ============
 
 * Your contribution here.
+* [#119](https://github.com/jenkinsci/ansicolor-plugin/pull/119): Add helper method to create AnsiHtmlOutputStream - [@jstrachan](https://github.com/jstrachan).
 * [#107](https://github.com/jenkinsci/ansicolor-plugin/pull/107): Removing startup banner - [@jglick](https://github.com/jglick).
 
 0.5.2 (08/17/2017)
--- a/src/main/java/hudson/plugins/ansicolor/AnsiColorConsoleLogFilter.java
+++ b/src/main/java/hudson/plugins/ansicolor/AnsiColorConsoleLogFilter.java
@@ -7,7 +7,6 @@ import hudson.model.AbstractBuild;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.io.Serializable;
-import java.util.logging.Level;
 import java.util.logging.Logger;
 
 /**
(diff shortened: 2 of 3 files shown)


dblock:
Forgive me if this is a dumb question, haven't been writing much Java recently ;) 

The helper seems to accomplish the same thing as the constructor, so why is this working vs. what existed before? I mean when you say "emitter stuff is not public", isn't `AnsiHtmlOutputStream.new` public?

If this is to be merged it needs a test for the publicly added API, please, and a line in CHANGELOG.

jstrachan:
@dblock ;) sorry I should have been more clear. The main reason for the refactor is:

* you cannot use `AnsiAttributeElement.Emitter` from any other package than `hudson.plugins.ansicolor` (as its not public) so we cannot create `AnsiHtmlOutputStream` anywhere else than from source code in the `hudson.plugins.ansicolor` package
* I wanted to reuse the default implementation of `AnsiAttributeElement.Emitter` that ansicolor uses

So this minor refactor and new method was the simplest thing I could think of to work around it really. 

The new code is still invoked by the current tests as the code is invoked by the constructor of `new LineTransformationOutputStream()` so there should not need to be any new tests to ensure coverage.

Lemme add a CHANGELOG entry now...

jstrachan:
here's the hack I had to do to work around the above issues btw https://github.com/jenkinsci/updatebot-plugin/blob/master/src/main/java/hudson/plugins/ansicolor/AnsiHelper.java#L17

i.e. a dirty hack putting some code into the `hudson.plugins.ansicolor` package and copy-pasting the implementation of `AnsiAttributeElement.Emitter()` - all of which I'd rather replace with the new static method

dblock:
Why don't you just make these attributes public?

jstrachan:
@dblock I was going for the smallest refactor that had the most minimal impact on the public API. i.e. I added 1 new static function call. 

Another approach could be:
* make `AnsiAttributeElement` and `AnsiAttributeElement. Emitter` public
* refactor this anonymous inner class to a new public class https://github.com/jenkinsci/ansicolor-plugin/blob/master/src/main/java/hudson/plugins/ansicolor/AnsiColorConsoleLogFilter.java#L38-L47

I went with the former as it had less impact on the public API and usually folks tend to explicitly opt out of making things `public` for a reason - so guessed you'd prefer the former. Happy to submit a PR for the latter if you'd prefer it though?

dblock:
I think the inner class can safely become a public class. Helper methods are "APIs coding for a client".

jglick:
Merge conflicts. If this is still relevant, best to recreate and show a downstream PR using the proposed new API.

PR #126, opened 2018-10-09
Use Java 8 on Travis

darxriggs:
This fixes the current build issue on Travis for Linux.

Currently Java 10 is the default on Travis but the project does not yet compile with Java 10. Specifying the desired JDK version is anyway a good idea.

The OSX build can also be fixed. But I would suggest to remove testing on OSX completely. I haven't seen it for other plugins. A plugin is anyway build centrally and provided in binary form. So testing the build on another platform doesn't really provide that much.

Another question is if the plugin should better be build using the Jenkins infrastructure - see https://github.com/jenkins-infra/pipeline-library and https://ci.jenkins.io/job/Plugins/.

Therefore I propose to replace the `.travis.yml` with `Jenkinsfile`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,8 @@
 language: java
-
+jdk:
+  - oraclejdk8
 matrix:
   include:
     - os: linux
     - os: osx
-      env: JAVA_HOME=$(/usr/libexec/java_home)
+      osx_image: xcode9.3


darxriggs:
The OSX build can also be fixed. But I would suggest to remove testing on OSX completely. I haven't seen it for other plugins. A plugin is anyway build centrally and provided in binary form. So testing the build on another platform doesn't really provide that much.

dblock:
Still fails with 

```
[ERROR] java.nio.file.NoSuchFileException: /Users/travis/build/jenkinsci/ansicolor-plugin/target/classes/META-INF/annotations/hudson.Extension
```

I am not too excited about using Jenkins infrastructure. I don't see why that should matter.

darxriggs:
Both Linux and OSX are ok now on Travis. So you can have a look again.

dblock:
Thanks for fixing this!

PR #127, opened 2018-10-11
Cleanup pom.xml

darxriggs:
Please have a look at the individual commits when reviewing.
One aspect is cleaned per commit.
In the last commit automated reordering of entries was performed, therefore looking only at the final diff could be a bit difficult.


dblock:
Amen

PR #128, opened 2018-10-30
[JENKINS-54133] Pregenerating console notes sent to agents for some common cases

jglick:
[JENKINS-54133](https://issues.jenkins-ci.org/browse/JENKINS-54133). Does not attempt to pregenerate notes for all cases, and in fact this seems to be impossible, though more common cases could certainly be added as needed.

As noted in JIRA, reimplementing the plugin as a `ConsoleAnnotatorFactory` would be a better solution in many respects, as well as fully solving this issue, but this would be a more ambitious refactoring than I am prepared to take on at the moment as it would involve inverting the control flow of `AnsiOutputStream`. See the corresponding Blue Ocean https://github.com/jenkinsci/blueocean-plugin/pull/1325 for a very general idea.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,8 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>2.25</version>
+    <version>3.25</version>
+    <relativePath />
   </parent>
 
   <artifactId>ansicolor</artifactId>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -45,23 +46,51 @@
   </distributionManagement>
 
   <properties>
-    <jenkins.version>1.642.3</jenkins.version>
+    <jenkins.version>2.121.2</jenkins.version>
```
jglick: According to [stats](http://stats.jenkins.io/pluginversions/ansicolor.html) this is not a big deal, even if there are active trunk releases of this plugin for unrelated changes.

But, if baseline is an obstacle, the POM changes could be reverted. In that case the only `src/main/` change would be to convert some uses of lambdas to inner classes, pretty simple; the `src/test/` code could remain, but would no longer reproduce the problem unless run under `plugin-compat-tester`.

src/main/java/hudson/plugins/ansicolor/AnsiColorConsoleLogFilter.java
```diff
@@ -20,10 +24,40 @@
...
 
     public AnsiColorConsoleLogFilter(AnsiColorMap colorMap) {
         super();
         this.colorMap = colorMap;
+        this.notes = new HashMap<>();
+        // some cases of AnsiHtmlOutputStream.setForegroundColor:
+        for (AnsiColorMap.Color color : AnsiColorMap.Color.values()) {
+            pregenerateNote(new AnsiAttributeElement(AnsiAttributeElement.AnsiAttrType.FG, "span", "style=\"color: " + colorMap.getNormal(color.ordinal()) + ";\""));
+        }
+        // TODO other cases, and other methods
```
jglick: The tedious part. Handling bright colors, `swapColors`, bold/italic/underline, etc. Basically every call to the `AnsiAttributeElement` constructor from `AnsiHtmlOutputStream` which could be predicted given an `AnsiColorMap`—probably all but the explicit use of RGB coördinates, which I am guessing are rarely used anyway.

src/main/java/hudson/plugins/ansicolor/AnsiColorConsoleLogFilter.java
```diff
@@ -20,10 +24,40 @@
...
+    }
+    
+    private void pregenerateNote(AnsiAttributeElement element) {
+        element.emitOpen(html -> pregenerateNote(html));
+        element.emitClose(html -> pregenerateNote(html));
+    }
+    
+    private void pregenerateNote(String html) {
+        if (!notes.containsKey(html)) {
+            JenkinsJVM.checkJenkinsJVM();
```
jglick: See https://github.com/jenkinsci/ant-plugin/pull/32 for a simpler patch that illustrates the idiom.

src/test/java/hudson/plugins/ansicolor/AnsiColorBuildWrapperTest.java
```diff
@@ -42,17 +45,21 @@ public void testDecorateLogger() {
...
         story.addStep(new Statement() {
 
             @Override
             public void evaluate() throws Throwable {
                 Assume.assumeTrue(!Functions.isWindows());
+                story.j.createSlave();
                 WorkflowJob p = story.j.jenkins.createProject(WorkflowJob.class, "p");
                 p.setDefinition(new CpsFlowDefinition(
-                        "node {\n"
+                        "node('!master') {\n"
```
dblock: Feels like this should be a separate test.

dblock:
@jglick We could also use another maintainer here, let us know if you're interested.

src/test/java/hudson/plugins/ansicolor/AnsiColorBuildWrapperTest.java
jglick: It seemed unnecessary as `AnsiColorStepTest` already tests the behavior on master.

jglick:
> We could also use another maintainer here, let us know if you're interested.

Sorry, I could not commit to that.

jglick:
> maybe it would be good to add [`AnsiAttrType.BOLD`] as well

Probably. I was reluctant to get into cases for which there is no automated test coverage. It is possible the relevant parts of `AnsiHtmlOutputStreamTest` could be reworked to provide coverage for the `AnsiColorConsoleLogFilter` constructor in a separate test suite, with most cases initially `@Ignore`d.

dblock:
Merged, thanks.

PR #130, opened 2018-11-01
Remote console note test coverage & fixes

jglick:
Provides the test coverage I proposed in #128, and fixes some straightforward cases that are now tested.

Subsumes #129, without which input text lacking a final newline would not be sent at all, unless `println` and `flush` were called, at which point the result string would contain an extra newline.


dblock:
@dwnusbaum You seem to know what you're doing and be working for cloudbees. Want to help out with the plugin a bit? I think we could use an active maintainer, maybe someone to make the next release.

dwnusbaum:
@dblock I can help you with the next release if needed, but unfortunately I cannot commit to being an active maintainer.

dblock:
#131

PR #132, opened 2018-11-05
Switching plugin to use ConsoleAnnotator rather than ConsoleNote

jglick:


src/test/java/hudson/plugins/ansicolor/AnsiHtmlOutputStreamTest.java
```diff
@@ -610,7 +610,7 @@ private void assertThatAnnotateIs(AnsiColorMap colorMap, String ansi, String htm
         assertThat(annotate(ansi, colorMap), is(html));
     }
 
-    protected String annotate(String text, AnsiColorMap colorMap) throws IOException {
+    private String annotate(String text, AnsiColorMap colorMap) throws IOException {
```
jglick: Reverting this bit from #130.

src/main/java/hudson/plugins/ansicolor/ColorConsoleAnnotator.java
```diff
@@ -0,0 +1,135 @@
...
+
+        @Override
+        public ConsoleAnnotator<Object> newInstance(Object context) {
+            LOGGER.fine("context=" + context);
+            if (context instanceof Run) {
+                ColorizedAction action = ((Run) context).getAction(ColorizedAction.class);
+                if (action != null) {
+                    return new ColorConsoleAnnotator(action.colorMapName, ((Run) context).getCharset().name());
+                }
+            } else if (Jenkins.get().getPlugin("workflow-api") != null && context instanceof FlowNode) {
```
jglick: Did not try to verify behavior when the plugin is not in fact installed. Generally speaking, `optional` deps are dangerous. I would rather just make it a hard dep.

src/main/java/hudson/plugins/ansicolor/ColorConsoleAnnotator.java
```diff
@@ -0,0 +1,135 @@
...
+    @Override
+    public ConsoleAnnotator<Object> annotate(Object context, MarkupText text) {
+        String s = text.getText();
+        if (s.indexOf('\u001B') != -1) {
+            AnsiColorMap colorMap = Jenkins.get().getDescriptorByType(AnsiColorBuildWrapper.DescriptorImpl.class).getColorMap(colorMapName);
+            CountingOutputStream outgoing = new CountingOutputStream(new NullOutputStream());
+            class EmitterImpl implements AnsiAttributeElement.Emitter {
+                CountingOutputStream incoming;
+                int adjustment;
+                int lastPoint = -1; // multiple HTML tags may be emitted for one control sequence
```
jglick: The triumph of test-driven development over comprehensibility.

pom.xml
```diff
@@ -46,7 +46,7 @@
   </distributionManagement>
 
   <properties>
-    <jenkins.version>2.121.2</jenkins.version>
+    <jenkins.version>2.145</jenkins.version> <!-- to pick up https://github.com/jenkinsci/jenkins/pull/3662 -->
```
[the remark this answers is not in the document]
dwnusbaum: Since that fix is not in any LTS line yet, I think it would be good to release what we currently have in master for anyone using JEP-210 in recent LTS lines, and then we can release this afterwards as a complete fix for future LTS lines.

dblock:
@dwnusbaum Can you code review/merge this please? I'm going to try to do less work here ;)

src/main/java/hudson/plugins/ansicolor/ColorConsoleAnnotator.java
dwnusbaum: I did some testing with the plugins not installed and things seem to work fine. New freestyle builds are still colored correctly, and executing `Jenkins.get().getPlugin("workflow-api")` in the script console returns null, so the short-circuiting behavior here looks fine to me.

src/test/java/hudson/plugins/ansicolor/AnsiColorBuildWrapperTest.java
```diff
@@ -46,7 +53,40 @@ public void testDecorateLogger() {
...
+                }
+            });
+            FreeStyleBuild b = r.buildAndAssertSuccess(p);
+            StringWriter writer = new StringWriter();
+            b.getLogText().writeHtmlTo(0L, writer);
+            String html = writer.toString();
+            System.out.print(html);
+            assertThat(html.replaceAll("<span style=\"display: none\">.+?</span>", ""),
+                allOf(
+                    containsString("[<b><span style=\"color: #1E90FF;\">INFO</span></b>]"),
```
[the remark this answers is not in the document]
dwnusbaum: A little awkward how some of the hidden ANSI escape sequences end up nested in other tags, but it works 🤷‍♂️ . In other words, I would've expected the inner span with ` style="display: none"` to come after the `</b>` tag when displaying `[INFO]` in a colorized Maven build:

    [
    <span style="display: none">\u001B[1;34m</span> 
    <b>
        <span style="color: #1E90FF;">
             INFO
             <span style="display: none">\u001B[m</span> 
        </span>
    </b>
    ]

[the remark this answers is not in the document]
jglick: Right, I decided it was not worth the bother to fix.

reinholdfuereder:
Thanks for the improvements/fixes: besides the aforementioned Jenkins issues I think this also really solves JENKINS-41200, JENKINS-46324 and JENKINS-43107

cmurphy208:
@jglick Does this cover the pipeline.log file that contains all the output during a pipeline?

Is there a way to completely strip the ansi codes for this pipeline.log file? (or replace them with html tags)?

jglick:
> Does this cover the pipeline.log file that contains all the output during a pipeline?

Not sure what exactly you are asking.

> Is there a way to completely strip the ansi codes for this pipeline.log file?

Only by displaying the `…/console` link the Jenkins GUI.

cmurphy208:
Just to clarify for all other parties:
I was talking about the pipeline/build log accessible via blue ocean UI (i.e. the pop-out button above the comment section). It appears that this was out of the scope of this change according @jglick 
 I intend to file a blue-ocean plugin bug on his behalf.
@jglick can you advice the preferred way to have a discussion with your team in the future?

jglick:
Yes Blue Ocean has its own handling of ANSI escapes which was never compatible with the `ConsoleNote`s formerly used in this plugin—one of the motivations for getting rid of the notes. If Blue Ocean does not process escapes in the full log view, that is a Blue Ocean issue.

PR #137, opened 2018-11-16
Allow escape sequences to span multiple lines and support default fg/bg colors

### LINKED ISSUE #135: sgr0 not supported anymore
Using version 0.5.3 of the AnsiColor plugin I could write a script such as:

bold=$(tput -Txterm bold)
norm=$(tput -Txterm sgr0)

echo "${bold}bold text${norm} and normal text"

After upgrading to 0.6.0, the bold text is still bold, but the normal text shows some stange characters followed by a text that is still bold

dwnusbaum:
Fixes #135.
 
This PR currently adds tests and fixes for 2 issues:

1. If an escape sequence is not closed at the end of a line, its tag cannot be closed and its style modifications will carry over to all subsequent lines.
    * This PR currently closes all tags that are open at the end of a line and then reopens them on the next line so that ANSI sequences in the line can close those tags correctly. I am not sure whether this is the desired behavior, or if we should instead close any tags that are open at the end of a line and leave them closed for good.
2. When using a color map with a default foreground or background, the default foreground and background are not used.
    * This PR handles default foreground and background settings. I am not sure how often default foreground/background colors are used, so maybe we don't care about this, but it was easy enough to add support for it.

### DIFF
--- a/src/main/java/hudson/plugins/ansicolor/AnsiAttributeElement.java
+++ b/src/main/java/hudson/plugins/ansicolor/AnsiAttributeElement.java
@@ -65,6 +65,11 @@ class AnsiAttributeElement {
         return result;
     }
 
+    @Override
+    public String toString() {
+        return "<" + name + " " + attributes + ">";
+    }
+
     public static AnsiAttributeElement bold() {
         return new AnsiAttributeElement(AnsiAttributeElement.AnsiAttrType.BOLD, "b", "");
     }
--- a/src/main/java/hudson/plugins/ansicolor/AnsiAttributeElement.java
+++ b/src/main/java/hudson/plugins/ansicolor/AnsiAttributeElement.java
@@ -67,7 +67,7 @@ class AnsiAttributeElement {
 
     @Override
     public String toString() {
-        return "<" + name + " " + attributes + ">";
+        return "AnsiAttributeElement{ansiAttrType=" + ansiAttrType + ",name=" + name + ",attributes=" + attributes + "}";
     }
 
     public static AnsiAttributeElement bold() {
--- a/src/main/java/hudson/plugins/ansicolor/AnsiAttributeElement.java
+++ b/src/main/java/hudson/plugins/ansicolor/AnsiAttributeElement.java
@@ -1,5 +1,7 @@
 package hudson.plugins.ansicolor;
 
+import java.io.Serializable;
+
 /**
  * Represents an HTML elements which maps to an ANSI attribute.
  *
@@ -12,7 +14,9 @@ package hudson.plugins.ansicolor;
  * other software or testing the HTML may be emitted otherwise.
  */
 
-class AnsiAttributeElement {
+class AnsiAttributeElement implements Serializable {
+    private static final long serialVersionUID = 1L;
+
     public static enum AnsiAttrType {
         DEFAULT, BOLD, ITALIC, UNDERLINE, STRIKEOUT, FRAMED, OVERLINE, FG, BG, FGBG
     }
--- a/src/main/java/hudson/plugins/ansicolor/AnsiHtmlOutputStream.java
+++ b/src/main/java/hudson/plugins/ansicolor/AnsiHtmlOutputStream.java
@@ -30,6 +30,7 @@ import hudson.util.NullStream;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.util.ArrayList;
+import java.util.List;
 import java.util.Stack;
 
 /**
@@ -93,12 +94,19 @@ public class AnsiHtmlOutputStream extends AnsiOutputStream {
         this.out = logOutput;
     }
 
-    private void openTag(AnsiAttributeElement tag) throws IOException {
+    /**
+     * @return A copy of the {@link AnsiAttributeElement}s which are currently opened, in order from outermost to innermost tag.
+     */
+    /*package*/ List<AnsiAttributeElement> getOpenTags() {
+        return new ArrayList<>(openTags);
+    }
+
+    /*package*/ void openTag(AnsiAttributeElement tag) throws IOException {
         openTags.add(tag);
         tag.emitOpen(emitter);
     }
 
-    private void closeOpenTags(AnsiAttrType until) throws IOException {
+    /*package*/ void closeOpenTags(AnsiAttrType until) throws IOException {
         while (!openTags.isEmpty()) {
             int index = openTags.size() - 1;
             if (until != null && openTags.get(index).ansiAttrType == until)
--- a/src/main/java/hudson/plugins/ansicolor/AnsiHtmlOutputStream.java
+++ b/src/main/java/hudson/plugins/ansicolor/AnsiHtmlOutputStream.java
@@ -106,7 +106,7 @@ public class AnsiHtmlOutputStream extends AnsiOutputStream {
         tag.emitOpen(emitter);
     }
 
-    /*package*/ void closeOpenTags(AnsiAttrType until) throws IOException {
+    private void closeOpenTags(AnsiAttrType until) throws IOException {
         while (!openTags.isEmpty()) {
             int index = openTags.size() - 1;
             if (until != null && openTags.get(index).ansiAttrType == until)
--- a/src/main/java/hudson/plugins/ansicolor/AnsiHtmlOutputStream.java
+++ b/src/main/java/hudson/plugins/ansicolor/AnsiHtmlOutputStream.java
@@ -30,8 +30,10 @@ import hudson.util.NullStream;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Stack;
+import javax.annotation.Nonnull;
 
 /**
  * Filters an outputstream of ANSI escape sequences and emits appropriate HTML elements instead.
@@ -60,16 +62,27 @@ public class AnsiHtmlOutputStream extends AnsiOutputStream {
     private boolean swapColors = false;  // true if negative / inverse mode is active (esc[7m)
 
     // A Deque might be a better choice, but we are constrained by the Java 5 API.
-    private ArrayList<AnsiAttributeElement> openTags = new ArrayList<AnsiAttributeElement>();
+    private ArrayList<AnsiAttributeElement> openTags;
 
     private OutputStream logOutput;
 
-    public AnsiHtmlOutputStream(final OutputStream os, final AnsiColorMap colorMap,
-        final AnsiAttributeElement.Emitter emitter) {
+    /**
+     * @param tagsToOpen A list of tags to open in the given order immediately after opening the tag for the default
+     * foreground/background colors (if such colors are specified by the color map) before any data is written to the
+     * underlying stream.
+     */
+    /*package*/ AnsiHtmlOutputStream(final OutputStream os, final AnsiColorMap colorMap,
+        final AnsiAttributeElement.Emitter emitter, @Nonnull List<AnsiAttributeElement> tagsToOpen) {
         super(os);
         this.logOutput = os;
         this.colorMap = colorMap;
         this.emitter = emitter;
+        this.openTags = new ArrayList<>(tagsToOpen);
+    }
+
+    public AnsiHtmlOutputStream(final OutputStream os, final AnsiColorMap colorMap,
+        final AnsiAttributeElement.Emitter emitter) {
+        this(os, colorMap, emitter, Collections.emptyList());
     }
 
     // Debug output for plugin developers. Puts the debug message into the html page
@@ -101,7 +114,7 @@ public class AnsiHtmlOutputStream extends AnsiOutputStream {
         return new ArrayList<>(openTags);
     }
 
-    /*package*/ void openTag(AnsiAttributeElement tag) throws IOException {
+    private void openTag(AnsiAttributeElement tag) throws IOException {
         openTags.add(tag);
         tag.emitOpen(emitter);
     }
@@ -206,6 +219,9 @@ public class AnsiHtmlOutputStream extends AnsiOutputStream {
         // the preamble is an ANSI escape sequence itself.
 
         if (state == State.INIT) {
+            List<AnsiAttributeElement> tagsToOpen = new ArrayList<>(openTags);
+            openTags.clear();
+
             Integer defaultFg = colorMap.getDefaultForeground();
             Integer defaultBg = colorMap.getDefaultBackground();
 
@@ -213,6 +229,10 @@ public class AnsiHtmlOutputStream extends AnsiOutputStream {
                 openTag(new AnsiAttributeElement(AnsiAttrType.DEFAULT, "div", "style=\"" +
                         (defaultBg != null ? "background-color: " + colorMap.getNormal(defaultBg) + ";" : "") +
                         (defaultFg != null ? "color: " + colorMap.getNormal(defaultFg) + ";" : "") + "\""));
+            }
+
+            for (AnsiAttributeElement tag : tagsToOpen) {
+                openTag(tag);
             }
 
             state = State.DATA;
--- a/src/main/java/hudson/plugins/ansicolor/ColorConsoleAnnotator.java
+++ b/src/main/java/hudson/plugins/ansicolor/ColorConsoleAnnotator.java
@@ -71,10 +71,10 @@ final class ColorConsoleAnnotator extends ConsoleAnnotator<Object> {
     public ConsoleAnnotator<Object> annotate(Object context, MarkupText text) {
         String s = text.getText();
         List<AnsiAttributeElement> nextOpenTags = openTags;
-        // TODO: As a performance improvement, we could create a branch for `s.indexOf('\u001B') == -1 & !openTags.isEmpty()`
-        // that just surrounds the text in the appropriate tags without going through AnsiHtmlOutputStream.
-        if (s.indexOf('\u001B') != -1 || !openTags.isEmpty()) {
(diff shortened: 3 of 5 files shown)


dblock:
This test seems to have passed ...

src/test/java/hudson/plugins/ansicolor/AnsiColorBuildWrapperTest.java
```diff
@@ -88,6 +89,34 @@ public boolean perform(AbstractBuild<?, ?> build, Launcher launcher, BuildListen
         });
     }
 
+    @Test
+    @Ignore
```
dwnusbaum: @dblock The test is ignored, so it never even ran.

dwnusbaum:
I am working on a fix for this which may help with [JENKINS-34019](https://issues.jenkins-ci.org/browse/JENKINS-34019) because the fix here will close open tags at the end of each line and reopen them on the next line, which seems to be the suggested approach to fix that ticket.

src/main/java/hudson/plugins/ansicolor/AnsiAttributeElement.java
```diff
@@ -65,6 +65,11 @@ public int hashCode() {
         return result;
     }
 
+    @Override
+    public String toString() {
+        return "<" + name + " " + attributes + ">";
```
dwnusbaum: Just for debugging, maybe better to switch to `AnsiAttributeElement name={}, attributes={}` so we don't emit valid HTML directly in logs or something.

src/main/java/hudson/plugins/ansicolor/AnsiHtmlOutputStream.java
```diff
@@ -93,12 +94,19 @@ private void stopConcealing() {
...
 
-    private void openTag(AnsiAttributeElement tag) throws IOException {
+    /**
+     * @return A copy of the {@link AnsiAttributeElement}s which are currently opened, in order from outermost to innermost tag.
+     */
+    /*package*/ List<AnsiAttributeElement> getOpenTags() {
+        return new ArrayList<>(openTags);
+    }
+
+    /*package*/ void openTag(AnsiAttributeElement tag) throws IOException {
```
dwnusbaum: Making these methods accessible seems bad. Maybe better to provide a constructor that takes tags to open and opens them as the first part of the state machine, and then just `close` the stream in `ColorConsoleAnnotator` once we are done with it so `closeOpenTags` doesn't need to be called externally. I will investigate.

src/main/java/hudson/plugins/ansicolor/ColorConsoleAnnotator.java
```diff
@@ -70,40 +80,67 @@
...
+                ansiOs.closeOpenTags(null);
             } catch (IOException x) {
                 LOGGER.log(Level.WARNING, null, x);
             }
             LOGGER.finer(() -> "\"" + StringEscapeUtils.escapeJava(s) + "\" → \"" + StringEscapeUtils.escapeJava(text.toString(true)) + "\"");
         }
-        return this;
+        return openTags == nextOpenTags
+                ? this
+                : new ColorConsoleAnnotator(colorMapName, charset, nextOpenTags);
```
dwnusbaum: Based on my reading of the `ConsoleAnnotator` Javadocs, if we want to track state from line-to-line then we need to handle it by returning a new instance like this. Maybe wrong to `return this` instead of creating a new state when the open tags match.

src/main/java/hudson/plugins/ansicolor/ColorConsoleAnnotator.java
```diff
@@ -70,40 +80,67 @@
...
+    }
+
+    private Object readResolve() {
```
dwnusbaum: The `ConsoleAnnotator` Javadocs mention serialization, so I added this, although it's not clear to me if serialized `ConsoleAnnotator` objects would survive a restart (in which case this method is needed) or only exist in memory or some other transient form (in which case this method can be deleted).

src/main/java/hudson/plugins/ansicolor/ColorConsoleAnnotator.java
```diff
@@ -51,17 +54,24 @@
...
+    ColorConsoleAnnotator(String colorMapName, String charset) {
+        this(colorMapName, charset, Collections.emptyList());
     }
 
     @Override
     public ConsoleAnnotator<Object> annotate(Object context, MarkupText text) {
         String s = text.getText();
-        if (s.indexOf('\u001B') != -1) {
+        List<AnsiAttributeElement> nextOpenTags = openTags;
+        if (s.indexOf('\u001B') != -1 || !openTags.isEmpty()) {
```
dwnusbaum: We could create a faster path for when `openTags.isEmpty && s.indexOf('\u001B') == -1`, since we'd just need to add tags at the beginning and end of the text without changing it. Not sure how important it is, but probably worth a comment in the source.

dwnusbaum: Added in a95ea81.

src/main/java/hudson/plugins/ansicolor/AnsiHtmlOutputStream.java
dwnusbaum: It was easy to call `close` directly which removed the need to call `closeOpenTags`, but `openTag` is trickier. 

If the tags can be passed to the constructor, and are opened at the beginning of `AnsiHtmlOutputStream#write`, then in `ColorConsoleOutputStream` we end up with an `IndexOutOfBoundsException` and malformed output because the `CountingInputStream` records the first byte written before the initial tags are opened, so the offsets are incorrect. Calling `openTags` directly keeps the `CountingInputStream` from updating, which keeps this from being a problem. However, I think this means that there is already a bug in the existing implementation for any color map that sets a default foreground or background, since [this call](https://github.com/jenkinsci/ansicolor-plugin/blob/39f63c28386f432c4ff17ea4de8f4f2ba21a2089/src/main/java/hudson/plugins/ansicolor/AnsiHtmlOutputStream.java#L205) in `AnsiHtmlOutputStream#write` will cause the first character in the output to be hidden incorrectly.

I will keep thinking about the best way to handle this.

src/main/java/hudson/plugins/ansicolor/AnsiAttributeElement.java
dwnusbaum: Done in a95ea81.

src/main/java/hudson/plugins/ansicolor/ColorConsoleAnnotator.java
jglick: I thought the suggested fix for JENKINS-34019 was just to close all outstanding tags at the end of a line, as if this were a user error, not to try to make them apply across lines.

dwnusbaum: > When a style continues to the next line, the ConsoleAnnotatorFactory can insert a closing tag at the end of the line and an opening tag at the beginning of the next line. This prevents the timestamps from being affected.

[Full comment](https://issues.jenkins-ci.org/browse/JENKINS-34019?focusedCommentId=272506&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-272506). What you propose would certainly be simpler, but I'm not sure what the expected behavior is. What do you think should be the behavior for [the test case I added](https://github.com/jenkinsci/ansicolor-plugin/blob/a95ea81c6d529e16da44437d1cf12b025390a4e2/src/test/java/hudson/plugins/ansicolor/AnsiColorBuildWrapperTest.java#L99)?

src/main/java/hudson/plugins/ansicolor/AnsiHtmlOutputStream.java
dwnusbaum: Thinking about this some more, since `AnsiAttributeElement` is already package-private, I don't think this is a big problem, but either way I switched to using the constructor for open tags in `2c19b61` so that I could also fix the issue with default foreground/background colors.

src/test/java/hudson/plugins/ansicolor/AnsiColorBuildWrapperTest.java
```diff
@@ -88,6 +88,61 @@ public boolean perform(AbstractBuild<?, ?> build, Launcher launcher, BuildListen
...
+                }
+            });
+            FreeStyleBuild b = r.buildAndAssertSuccess(p);
+            StringWriter writer = new StringWriter();
+            b.getLogText().writeHtmlTo(0L, writer);
+            String html = writer.toString();
+            System.out.print(html);
+            assertThat(html.replaceAll("<span style=\"display: none\">.+?</span>", ""),
+                allOf(
+                    containsString("<b><span style=\"color: #1E90FF;\">This text should be bold and blue\n</span></b>"),
```
dwnusbaum: The literal `\n` characters in these tests would probably cause them to fail on Windows.

src/test/java/hudson/plugins/ansicolor/AnsiColorBuildWrapperTest.java
```diff
@@ -88,6 +88,61 @@ public boolean perform(AbstractBuild<?, ?> build, Launcher launcher, BuildListen
         });
     }
 
+    @Test
```
dwnusbaum: I should create a JIRA issue for these and add `@Issue`.

dwnusbaum:
Needs to be merged with master after the change to use comments instead of `display: none` in #142.

src/main/java/hudson/plugins/ansicolor/ColorConsoleAnnotator.java
```diff
@@ -51,59 +54,107 @@
...
+            // correct in case those tags are closed in the middle of this line.
+            try (AnsiHtmlOutputStream ansiOs = new AnsiHtmlOutputStream(outgoing, colorMap, emitter, openTags);
+                    CountingOutputStream incoming = new CountingOutputStream(ansiOs)) {
+                emitter.incoming = incoming;
+                // We need to write one UTF-16 code unit at a time so that byte offsets match Java character offsets when inserting HTML.
+                for (int i = 0; i < s.length(); i++) {
+                    // TODO: Do we need to switch to using code points to support Unicode characters outside of the BMP?
+                    char c = s.charAt(i);
+                    if ((c & 0xFF00) != 0) {
+                        incoming.write(new byte[] { (byte) ((c & 0xFF00) >> 1), (byte) (c & 0x00FF) });
```
dwnusbaum: `(byte) (c & 0x00FF)` is equivalent to `(byte) c`, maybe better to switch, although the explicitness might help future readers.

dwnusbaum: Also I guess this is totally wrong for some charsets, working on fixing it.

dwnusbaum: Fixed by c1d7aab.

dblock:
@jglick Can I leave this to you to review/merge/comment?

src/main/java/hudson/plugins/ansicolor/ColorConsoleAnnotator.java
jglick: I feel like it does not matter much, since it is unlikely people would have multiline formatting intentionally.

src/main/java/hudson/plugins/ansicolor/ColorConsoleAnnotator.java
jglick: Only in a browser, so I think this could be dropped.

src/main/java/hudson/plugins/ansicolor/ColorConsoleAnnotator.java
```diff
@@ -51,59 +54,102 @@
...
-                byte[] data = s.getBytes(charset);
-                for (int i = 0; i < data.length; i++) {
-                    // Do not use write(byte[]) as offsets in incoming would not be accurate.
-                    incoming.write(data[i]);
+            // We need to reopen tags that were still open at the end of the previous line so the stream's state is
+            // correct in case those tags are closed in the middle of this line.
+            try (AnsiHtmlOutputStream ansiOs = new AnsiHtmlOutputStream(outgoing, colorMap, emitter, openTags);
+                    CountingOutputStream incoming = new CountingOutputStream(ansiOs)) {
+                emitter.incoming = incoming;
+                // We need to write one UTF-16 code unit at a time so that byte offsets match Java character offsets when inserting HTML.
```
jglick: Needs to be resolved in favor of #143, right?

src/main/java/hudson/plugins/ansicolor/ColorConsoleAnnotator.java
dwnusbaum: Removed in 31bb937.

src/main/java/hudson/plugins/ansicolor/ColorConsoleAnnotator.java
dwnusbaum: #145 shows a problem with output of `terraform` at the command line, and based on a screenshot of the text it looks like it might be using multiline formatting, but I'm not sure. I'll leave this as-is for now until I have time to investigate in more depth.

jglick: Confirmed via

```sh
docker run --rm hashicorp/terraform plan 2>&1 | od -a
```

(Though it also looks like the Terraform output is buggy—ANSI sequences are not nestable, and it is emitting several `esc   [   0   m` sequences in a row.)

src/main/java/hudson/plugins/ansicolor/ColorConsoleAnnotator.java
```diff
@@ -50,39 +53,62 @@
...
 
     private final String colorMapName;
+    private final @Nonnull List<AnsiAttributeElement> openTags;
 
-    ColorConsoleAnnotator(String colorMapName) {
+    ColorConsoleAnnotator(String colorMapName, List<AnsiAttributeElement> openTags) {
         this.colorMapName = colorMapName;
-        LOGGER.fine("creating annotator with colorMapName=" + colorMapName);
+        this.openTags = openTags;
+        LOGGER.log(Level.FINE, "creating annotator with colorMapName={0} openTags={2}", new Object[] { colorMapName, openTags });
```
jglick: `{1}` not `{2}`

src/main/java/hudson/plugins/ansicolor/ColorConsoleAnnotator.java
```diff
@@ -50,39 +53,62 @@
...
     }
 
     @Override
     public ConsoleAnnotator<Object> annotate(Object context, MarkupText text) {
         String s = text.getText();
-        if (s.indexOf('\u001B') != -1) {
-            AnsiColorMap colorMap = Jenkins.get().getDescriptorByType(AnsiColorBuildWrapper.DescriptorImpl.class).getColorMap(colorMapName);
+        List<AnsiAttributeElement> nextOpenTags = openTags;
+        // TODO: As a performance improvement, we could create a branch where `s.indexOf('\u001B') == -1` but other
+        // conditions are true that surrounds the text in the appropriate tags without going through AnsiHtmlOutputStream.
```
jglick: Unnecessary—this case is likely quite rare.

src/main/java/hudson/plugins/ansicolor/ColorConsoleAnnotator.java
```diff
@@ -50,39 +53,62 @@
...
-                        lastPoint = incoming.getCount();
-                        int hide = incoming.getCount() - outgoing.getCount() - adjustment;
-                        LOGGER.log(Level.FINEST, "hiding {0} @{1}", new Object[] { hide, outgoing.getCount() + adjustment });
-                        text.addMarkup(outgoing.getCount() + adjustment, outgoing.getCount() + adjustment + hide, "<!--", "-->");
-                        adjustment += hide;
+                    int inCount = incoming.getCount();
+                    int outCount = outgoing.getCount() + adjustment;
+                    // All ANSI escapes sequences contain at least 2 bytes on modern platforms, so any HTML emitted
+                    // directly after the first character is received is due to the initialization process of the stream and
+                    // belongs at position 0 (i.e. default background/foreground colors).
```
jglick: I think I need more caffeine to understand this. :-)

src/main/java/hudson/plugins/ansicolor/ColorConsoleAnnotator.java
```diff
@@ -102,20 +128,29 @@ public void emitHtml(String html) {
                     }
                     incoming.write(c);
                 }
+                nextOpenTags = ansiOs.getOpenTags();
+                if (colorMap.getDefaultBackground() != null || colorMap.getDefaultForeground() != null) {
+                    // The default color scheme will be opened automatically at the beginning of the stream on the next
+                    // line, so we don't want to duplicate it.
+                    nextOpenTags.remove(0);
```
jglick: `.subList(1)` might be less scary to readers, though I see that the getter does make a fresh copy.

dwnusbaum: Good point. I tried switching to `sublist`, but the implementation is not serializable so it would require another copy, which seems excessive, so I added a comment to help clarify for future readers but left the original code.

src/main/java/hudson/plugins/ansicolor/ColorConsoleAnnotator.java
dwnusbaum: Yeah this code was and still is quite confusing to me. Thankfully, the `if` statement is covered by the newly added tests, and if it is removed the tests fail, so we should be able to avoid future regressions.

PR #147, opened 2019-01-10
[JENKINS-55139] Handle redundant occurrences of SGR0 in logs

### LINKED ISSUE #145: Control characters appearing in job output in 0.6.0
I'm seeing control characters show up in coloured job output where previously there were none before. It seems that this could be somewhat related to #141.

### Test environment

- Jenkins Version: 2.148
- AnsiColor Version: 0.6.0
- Hostsystem: Debian stretch

### Expected behavior

Up until 0.6.0 I would only occasionally see `[0m` characters show up and they did not affect output. The following appears when testing against 0.5.3

screen shot 2018-12-20 at 16 22 06

### Actual behavior

With 0.6.0 I now see more control characters and in some cases no colour output at all: 

screen shot 2018-12-20 at 16 08 49

In more complex jobs with more varied colour output, the colour output will not be properly bounded and will leak into following lines. 

The above outputs were reproduced by running the following job using the pipeline library: 
```
def label = "terraform-${UUID.randomUUID().toString()}"

podTemplate(label: label, inheritFrom: 'jnlp', containers: [
  containerTemplate(
      name: 'terraform',
      image: 'hashicorp/terraform',
      ttyEnabled: true,
      command: 'cat',
      alwaysPullImage: true,
  )
]) {
    node(label) {
        container('terraform'){
            ansiColor('xterm') {
                sh """
                 terraform plan
                """
            }
        }
    }
}
```

dwnusbaum:
Fixes #145 (specifically variant 2 noted in https://github.com/jenkinsci/ansicolor-plugin/issues/145#issuecomment-452858635). See also [JENKINS-55139](https://issues.jenkins-ci.org/browse/JENKINS-55139).

Some tools, such as Terraform and Ansible, emit SGR0 (`ESC [ 0 m`) ANSI sequences even when display settings have not been changed since the last occurrence of SGR0. Since display settings are unchanged, no HTML is emitted, so our way of hiding control sequences using MarkupText does not work correctly. In 0.5.x and earlier, this wasn't a problem because all ANSI sequences were filtered from the underlying stream unconditionally, but now we ignore the output of the stream, so we have to account for these redundant sequences. We could try to fix those tools specifically, but if at least 2 tools are doing this I would not be surprised if it is more common in general, so it seems reasonable to handle it more gracefully here.

My approach to fix the issue was to add a new method to `AnsiAttributeElement.Emitter` used specifically for this purpose. Maybe it would be better just call `emitter.emitHtml(null)` directly in `AnsiHtmlOutputStream` and add documentation to that method that the argument may be `null` under some circumstances (Although I think [updatebot-plugin](https://github.com/jenkinsci/updatebot-plugin/blob/4c8db79c112b670320b23596b01382f17a2de94c/src/main/java/hudson/plugins/ansicolor/AnsiHelper.java) would break if we did so, maybe we could switch to using the empty string instead to avoid that problem).

What text similar to the `testRedundantResets` test case looks like before the change (note that `info` has been hidden!):
screen shot 2019-01-09 at 15 04 12

and after:
screen shot 2019-01-09 at 15 07 03

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 
 * Your contribution here.
 * [#137](https://github.com/jenkinsci/ansicolor-plugin/pull/137): Allow escape sequences to span multiple lines and support color maps that set default background/foreground colors - [@dwnusbaum](https://github.com/dwnusbaum).
+* [#147](https://github.com/jenkinsci/ansicolor-plugin/pull/147): Handle redundant occurrences of SGR0 in build logs - [@dwnusbaum](https://github.com/dwnusbaum).
 
 0.6.1 (01/04/2019)
 ============
--- a/src/main/java/hudson/plugins/ansicolor/AnsiAttributeElement.java
+++ b/src/main/java/hudson/plugins/ansicolor/AnsiAttributeElement.java
@@ -28,6 +28,12 @@ class AnsiAttributeElement implements Serializable {
 
     public static interface Emitter {
         public void emitHtml(String html);
+        /**
+         * Called when SGR0 (Set Graphics Rendition 0) reset is encountered in the stream, but no tags have been opened
+         * since its last occurrence. If you are using the output of AnsiOutputStream directly, you probably don't need
+         * to implement this method since the escape sequence will be transparently filtered from the output.
+         */
+        default void emitRedundantReset() { }
     }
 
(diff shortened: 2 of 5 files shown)


dblock:
Looks good, CHANGELOG & Al needed, feel free to merge.

dwnusbaum:
Ok, I will try to merge+release later this week when I have some time.

mingfang:
confirming fix! thank you very much.

PR #159, opened 2019-06-13
Bump parent pom version

rpionke:
Changelog can be found here: https://github.com/jenkinsci/plugin-pom/releases


basil:
@rpionke I think this can be closed in favor of #161.

PR #165, opened 2019-09-20
Allow global setting

### LINKED ISSUE #2: Allow global setting
The plugin is very useful. The only thing that's nagging me is, that I have to enable it in every job. Given I have > 100 Jobs, this gets quite cumbersome ;)

What do you think about a global switch for enabling it per default?

kcboschert:
This allows for configuring a global color map name.

Screen Shot 2019-09-20 at 11 13 55 AM

This resolves #2.


dblock:
LGTM, merged

PR #167, opened 2020-02-12
Allow tests to be executed independently of the OS

tszmytka:
Two of the tests use expected result strings with hard-coded `\n` as line separators. Jenkins, on the other hand, consults the host OS while generating output. As a result the test cases fail when run on a OS that doesn't use `\n` as a line separator (like Windows).

This fixes the issue and allows running tests in a OS-agnostic way.

### DIFF
--- a/src/test/java/hudson/plugins/ansicolor/AnsiColorBuildWrapperTest.java
+++ b/src/test/java/hudson/plugins/ansicolor/AnsiColorBuildWrapperTest.java
@@ -107,11 +107,15 @@ public class AnsiColorBuildWrapperTest {
             b.getLogText().writeHtmlTo(0L, writer);
             String html = writer.toString();
             System.out.print(html);
-            assertThat(html.replaceAll("<!--.+?-->", ""),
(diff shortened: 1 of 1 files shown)


dblock:
Nice. Want to setup AppVeyor or some other Windows-based CI here?

tszmytka:
> 
> 
> Nice. Want to setup AppVeyor or some other Windows-based CI here?

I almost started to write that Travis CI [supports Windows builds](https://docs.travis-ci.com/user/reference/windows/), but then I read further that Java is not among the supported languages. I can see there are people hacking around that but I don't know if it is worth the effort.

Haven't used AppVeyor yet but I can give it a try.

dblock:
> Haven't used AppVeyor yet but I can give it a try.

I've used it before and it works fine. 

https://github.com/waffle/waffle and https://github.com/java-native-access/jna for examples.

tszmytka:
It's a shame that there's no jenkins instance for all the plugins to test/build on

PR #168, opened 2020-02-12
Work with moving sequences

tszmytka:
Currently the plugin incorrectly handles control sequences like:
- `CUU` Cursor Up
- `CUD` Cursor Down
- `CUF` Cursor Forward
- `CUB` Cursor Back
- `EL` Erase in Line

Using many popular tools in a job, like `docker`, produces output similar to:
```
+ docker-compose down
Stopping container_1               ... 
Stopping container_2               ... 
Stopping container_3               ... 
[4A[2K
Stopping container_1               done
[4B[2A[2K
```

This PR contains a fix - the mentioned sequences will not produce any output - as well as a test case against regression.
Merging this may very well fix https://github.com/jenkinsci/ansicolor-plugin/issues/151 as the actual problem looks very similar.

### DIFF
--- a/src/main/java/hudson/plugins/ansicolor/AnsiAttributeElement.java
+++ b/src/main/java/hudson/plugins/ansicolor/AnsiAttributeElement.java
@@ -1,23 +1,22 @@
 package hudson.plugins.ansicolor;
 
+import javax.annotation.Nonnull;
 import java.io.Serializable;
 
 /**
  * Represents an HTML elements which maps to an ANSI attribute.
- *
- * An ANSI attribute change may open a new element or close an earlier one (e.g. "bold" opens, "bold off" closes).
- * Because HTML elements cannot overlap, any other enclosing element must be closed beforehand and subsequently
- * reopened.
- *
- * How the HTML is actually emitted depends on the specified {@link AnsiAttributeElement.Emitter}.
- * For Jenkins, the Emitter creates {@link hudson.console.ConsoleNote}s as part of the stream, but for
+ * <p>
+ * An ANSI attribute change may open a new element or close an earlier one (e.g. "bold" opens, "bold off" closes). Because HTML elements cannot overlap, any other enclosing element must be closed
+ * beforehand and subsequently reopened.
+ * <p>
+ * How the HTML is actually emitted depends on the specified {@link AnsiAttributeElement.Emitter}. For Jenkins, the Emitter creates {@link hudson.console.ConsoleNote}s as part of the stream, but for
  * other software or testing the HTML may be emitted otherwise.
  */
 
 class AnsiAttributeElement implements Serializable {
     private static final long serialVersionUID = 1L;
 
-    public static enum AnsiAttrType {
+    public enum AnsiAttrType {
         DEFAULT, BOLD, ITALIC, UNDERLINE, STRIKEOUT, FRAMED, OVERLINE, FG, BG, FGBG
     }
 
@@ -26,14 +25,20 @@ class AnsiAttributeElement implements Serializable {
     String name;
     String attributes;
 
-    public static interface Emitter {
-        public void emitHtml(String html);
+    public interface Emitter {
+        void emitHtml(@Nonnull String html);
+
         /**
-         * Called when SGR0 (Set Graphics Rendition 0) reset is encountered in the stream, but no tags have been opened
-         * since its last occurrence. If you are using the output of AnsiOutputStream directly, you probably don't need
-         * to implement this method since the escape sequence will be transparently filtered from the output.
+         * Emit invisible ANSI sequence
+         * <p>
+         * Used for emitting output for sequences like:
+         * <li>SGR0 (Set Graphics Rendition 0)
+         * <li>CSI (Control Sequence Introducer)
+         * <p>
+         * In particular called when SGR0 reset is encountered in the stream, but no tags have been opened since its last occurrence. If you are using the output of AnsiOutputStream directly, you
+         * probably don't need to implement this method since the escape sequence will be transparently filtered from the output.
          */
-        default void emitRedundantReset() { }
+        default void emitInvisibleSequence() { }
     }
 
     public AnsiAttributeElement(AnsiAttrType ansiAttrType, String name, String attributes) {
@@ -106,7 +111,6 @@ class AnsiAttributeElement implements Serializable {
 
     public static AnsiAttributeElement overline() {
         return new AnsiAttributeElement(AnsiAttrType.OVERLINE, "span", "style=\"text-decoration: overline;\"");
-        // return new AnsiAttributeElement(AnsiAttrType.OVERLINE, "span", "style=\"border-top: 1px solid;\""); // alternate approach
     }
 
 }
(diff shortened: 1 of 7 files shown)


src/main/java/hudson/plugins/ansicolor/AnsiHtmlOutputStream.java
```diff
@@ -627,4 +627,33 @@ protected void processDefaultTextColor() throws IOException {
...
+    }
+
+    @Override
+    protected void processCursorUpLine(int count) throws IOException {
+        emitNothing();
+    }
+
+    private void emitNothing() {
+        emitter.emitHtml("");
+    }
```
dwnusbaum: Instead of adding this method here, I would rename [`ColorConsoleWrapper$EmitterImpl.emitRedundantReset()`](https://github.com/jenkinsci/ansicolor-plugin/blob/bb959c39fdfeef15f52c876df3a361c7468fef6c/src/main/java/hudson/plugins/ansicolor/ColorConsoleAnnotator.java#L105-L108) to `emitInvisibleAnsiSequence` or something similar and call it in the body of the methods here instead of adding a helper method. That that method emits `null` instead of `""`, which has [special handling](https://github.com/jenkinsci/ansicolor-plugin/blob/bb959c39fdfeef15f52c876df3a361c7468fef6c/src/main/java/hudson/plugins/ansicolor/ColorConsoleAnnotator.java#L89) inside of `emitHtml`, so as long as it works, I think it would be better to emit `null` to make it clear that nothing is actually being emitted.

tszmytka: > 
> 
> Instead of adding this method here, I would rename [`ColorConsoleWrapper$EmitterImpl.emitRedundantReset()`](https://github.com/jenkinsci/ansicolor-plugin/blob/bb959c39fdfeef15f52c876df3a361c7468fef6c/src/main/java/hudson/plugins/ansicolor/ColorConsoleAnnotator.java#L105-L108) to `emitInvisibleAnsiSequence` or something similar and call it in the body of the methods here instead of adding a helper method.

I have nothing against this part. 

> 
> 
> That method emits `null` instead of `""`, which has [special handling](https://github.com/jenkinsci/ansicolor-plugin/blob/bb959c39fdfeef15f52c876df3a361c7468fef6c/src/main/java/hudson/plugins/ansicolor/ColorConsoleAnnotator.java#L89) inside of `emitHtml`, so as long as it works, I think it would be better to emit `null` to make it clear that nothing is actually being emitted.

But I can't say that I agree with

> better to emit `null`

Explicitly providing a `null` as a method parameter is just a `NPE` waiting to happen. I specifically avoided calling `ColorConsoleWrapper$EmitterImpl.emitRedundantReset()` for this exact reason. Maybe a cleaner approach would be to define a private constant inside of `ColorConsoleWrapper$EmitterImpl`
```java
private static final String NO_CORR_TAGS = "";
```
use it in `ColorConsoleWrapper$EmitterImpl.emitRedundantReset()`
and then check for it in the condition
```java
if (!NO_CORR_TAGS.equals(html)) {
// ...
}
```

dwnusbaum: > Explicitly providing a null as a method parameter is just a NPE waiting to happen

You could add `@CheckForNull` to the `String html` parameter in `EmitterImpl.emitHtml` to get SpotBugs to warn about any issues if you are concerned about someone introducing an unsafe change, but we already have tests that pass `null` via `emitRedundantReset`, so if someone did make a change that didn't handle `null` correctly, the tests should start failing, and the issue would have to be fixed before the PR was merged.

If you really don't like that, I think it would work fine to switch `emitRedundantReset` to use `""`, and change the logic in `emitHtml` to compare `html` against `""` where it currently compares `html` to `null`, but that seems riskier in the sense that someone might break things in the future in a subtle way that goes unnoticed at first because `""` is a valid `String`.

tszmytka:
I've provided additional adjustments so closing the original conversation.

- All `AnsiOutputStream` methods are overridden
- Made them all call `AnsiAttributeElement.Emitter#emitInvisibleSequence` previously `emitRedundantReset`
- Restructured `EmitterImpl` to not rely on any special values (`null` or otherwise) and equipped the method param with `@Nonnull`. Now there is no way of anyone providing `""`, `null` or anything else and triggering the _special behavior_.
I really don't like that guerilla-style `EmitterImpl` class defined out of nowhere in the middle of a method but changing that would require a broader refactoring.

- Added tests

@dwnusbaum what do you think?

src/main/java/hudson/plugins/ansicolor/AnsiAttributeElement.java
```diff
@@ -26,14 +25,20 @@
...
+         * <p>
+         * Used for emitting output for sequences like:
+         * <li>SGR0 (Set Graphics Rendition 0)
+         * <li>CSI (Control Sequence Introducer)
+         * <p>
+         * In particular called when SGR0 reset is encountered in the stream, but no tags have been opened since its last occurrence. If you are using the output of AnsiOutputStream directly, you
+         * probably don't need to implement this method since the escape sequence will be transparently filtered from the output.
          */
-        default void emitRedundantReset() { }
+        default void emitInvisibleSequence() { }
```
dwnusbaum: Note: This breaks binary compatibility. We could avoid it by keeping `emitRedundantReset` and having it call `emitInvisibleSequence` internally, but [it doesn't look like `emitRedundantReset` was ever used by any other plugins](https://github.com/search?utf8=%E2%9C%93&q=user%3Ajenkinsci+emitRedundantReset&type=Code), so I think the change is fine as-is.

tszmytka: > 
> 
> Note: This breaks binary compatibility.

Are there other plugins extending `ansicolor-plugin`?

dwnusbaum: [updatebot plugin](https://github.com/jenkinsci/updatebot-plugin/blob/4c8db79c112b670320b23596b01382f17a2de94c/src/main/java/hudson/plugins/ansicolor/AnsiHelper.java#L34) implements `AnsiAttributeElement.Emitter` (the usage looks weird, it probably _shouldn't_ be extending ansicolor), but the plugin looks obsolete and [has less than 50 installs](https://plugins.jenkins.io/updatebot/), so I think it can be ignored.

tszmytka:
Do I need to do anything else in order to merge this?

dwnusbaum:
@tszmytka I just merged #169, can you resolve the conflict? Once we have a passing CI build, I am happy to merge and release in the next few days (CC @dblock as a heads up).

Could you also update CHANGELOG.md in this PR with your changes?

tszmytka:
> 
> 
> @tszmytka I just merged #169, can you resolve the conflict? Once we have a passing CI build, I am happy to merge and release in the next few days (CC @dblock as a heads up).
> 
> Could you also update CHANGELOG.md in this PR with your changes?

@dwnusbaum  All done. Thanks for the help.

dwnusbaum:
@tszmytka Thanks for fixing the bug!

PR #169, opened 2020-02-16
Provide a basic Jenkinsfile to utilize testing on jenkins

tszmytka:
This PR allows for running tests on [ci.jenkins.io](http://ci.jenkins.io).
The used `buildPlugin.recommendedConfigurations()` contain (among others) 
`[ platform: "windows", jdk: "8", jenkins: recentLTS, javaLevel: "8" ],` so a separate CI service (AppVeyor or any other) would not be needed.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>3.48</version>
+    <version>3.56</version>
     <relativePath />
   </parent>
 
@@ -42,7 +42,7 @@
   <properties>
     <revision>0.6.3</revision>
     <changelist>-SNAPSHOT</changelist>
-    <jenkins.version>2.145</jenkins.version> <!-- to pick up https://github.com/jenkinsci/jenkins/pull/3662 -->
+    <jenkins.version>2.164.3</jenkins.version> <!-- to pick up https://github.com/jenkinsci/jenkins/pull/3662 -->
     <java.level>8</java.level>
   </properties>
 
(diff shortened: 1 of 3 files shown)


dblock:
Can we set this up along with the PR to see it work? I can help if something needs to be turned on by a (co)maintainer.

dblock:
Also add badges or whatever is else necessary for this build to be visible, please.

tszmytka:
> 
> 
> Can we set this up along with the PR to see it work? I can help if something needs to be turned on by a (co)maintainer.

The instance is configured to periodically scan branches and PRs so I needed to provide at least a PR in order to "trigger" it. From what I can [see](https://ci.jenkins.io/job/Plugins/computation/consoleFull) it should pick everything up simply if there is a `Jenkinsfile` available and the last time it ran was `Thu Feb 06 18:26:00 UTC 2020`. I am hoping it will scan us again in a couple of days. 

For the moment the PR can remain unmerged. I will monitor the situation and once it actually executes anything I will check if it needs anything else and report back.

dwnusbaum:
I suspect that the Jenkinsfile in this PR will be ignored when the scan runs because @tszmytka does not have write permission on the repo. I'll refile the PR just in case that is happening, and once I do that, I think that further commits to this PR will be built as long as there are no changes to the Jenkinsfile after 89b5b1a and 89b5b1a is not force-pushed over in this PR.

dwnusbaum:
Ok, #170 is being built on Travis and ci.jenkins.io, so I am going to try closing and reopening this PR to see if that triggers a build on ci.jenkins.io.

dwnusbaum:
Ok, looks like that worked! The Jenkinsfile looks fine to me assuming all of the tests pass.

tszmytka:
@dwnusbaum You were right. 
The build runs but there is 1 test failing. This is strange as all tests now run fine on my machine (Win10 + JDK 8, 11 and 13) when run directly through the IDE (Intellij IDEA).
 
Tried to execute them like Jenkins - with `maven` and voilà - it fails. I will try to approach it later but for the time being I've disabled it on Windows.

I've pushed my commit but ci.jenkins.io still ignores it.
@dwnusbaum Could you please pull my branch into yours and help Jenkins again?

jglick:
I think you need to also update `workflow-support` to work on Java 11?

jglick:
Also the dependencies here are pretty old. I would suggest switching the Jenkins baseline to an LTS, and then using https://github.com/jenkinsci/bom for as many plugin dependencies as possible.

tszmytka:
OK, I found the problem with the test failing on win+maven. It was `surefire`s fault.
@dwnusbaum Can you please pull my branch again?
@jglick Thanks for the hint - I will try to update what is possible.

pom.xml
```diff
@@ -111,4 +111,15 @@
       <url>https://repo.jenkins-ci.org/public/</url>
     </pluginRepository>
   </pluginRepositories>
+
+  <build>
+    <plugins>
+      <plugin>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <configuration>
+          <argLine>-Dfile.encoding=${project.build.sourceEncoding}</argLine>
```
jglick: If you had to do this, I suspect the test is at fault.

tszmytka: No, this really is a Surefire bug: [Surefire never outputs UTF-8 under Windows](https://issues.apache.org/jira/browse/SUREFIRE-1220)
I can reproduce it all with the attached sample project up to version `3.0.0-M3` which ansicolor-plugin currently uses. 

I will see if this can also be dealt with simply by updating Surefire.

tszmytka:
I've added the newest Jenkins BOM and let it take care of the dependencies. All tests run on my Win10 with JDK8 and JDK11.

@dwnusbaum Could you please pull my branch into yours?

pom.xml
jglick: That is about test _output_. I am talking about the fact that the test apparently _passes or fails_ depending on system encoding.

jglick: Freestyle projects are unduly sensitive to master encoding. Pipeline projects as of [JEP-206](https://jenkins.io/jep/206) are not.

jglick: Anyway, I suppose there is not an easy way to make the test behave reproducibly without either using Pipeline (which this wrapper is not compatible with), or running all tests in UTF-8 encoding (as you did).

pom.xml
```diff
@@ -42,59 +42,50 @@
   <properties>
     <revision>0.6.3</revision>
     <changelist>-SNAPSHOT</changelist>
-    <jenkins.version>2.145</jenkins.version> <!-- to pick up https://github.com/jenkinsci/jenkins/pull/3662 -->
+    <jenkins.version>2.164.3</jenkins.version> <!-- to pick up https://github.com/jenkinsci/jenkins/pull/3662 -->
```
dwnusbaum: nit: The comment is unnecessary now that we are on an LTS baseline.

```suggestion
    <jenkins.version>2.164.3</jenkins.version>
```

PR #170, opened 2020-02-17
[DO NOT MERGE] Copy of #169 from a user with write permission on the repo

dwnusbaum:
Just refiling https://github.com/jenkinsci/ansicolor-plugin/pull/169 in case the Jenkinsfile in that PR is being ignored because @tszmytka does not have write permission on the repo. With this PR up, changes to the other PR should be built on ci.jenkins.io.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,1 @@
+buildPlugin(configurations: buildPlugin.recommendedConfigurations())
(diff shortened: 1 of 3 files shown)


dwnusbaum:
> I think you need to also update workflow-support to work on Java 11?

Please comment over in #170, this PR is just so that the other PR gets built.

PR #171, opened 2020-02-18
Always run ColorConsoleAnnotator but deactivate it unless and until a colormap is available

### LINKED ISSUE #136: Colors are not displayed immediately after upgrading to 0.6.0
### Test environment

- Jenkins Version: 2.151
- AnsiColor Version: 0.6.0
- Hostsystem: Linux

### Expected behavior

When a pipeline build is triggered, if I immediately open the logs as soon as I trigger `Build Now`, the colors for that build won't appear while I am tailing the logs on Jenkins.
If I press back on the browser, open the same build logs again, the logs appear in colors.
I am using a declarative pipeline and using `ansiColor('xterm')` inside `options`.

### Actual behavior

The colors should appear always. Reverting the plugin to 0.5.3 fixes the issue.

Create a pipeline with `ansiColor('xterm')` inside `options`.
Press build now and go fast to Console output. I believe you need to get there before the `ansiColor` option is executed.
The logs won't display colors anymore. It will display again if you go back and open logs a second time.

jglick:
Fixes #136.

### DIFF
--- a/src/main/java/hudson/plugins/ansicolor/ColorConsoleAnnotator.java
+++ b/src/main/java/hudson/plugins/ansicolor/ColorConsoleAnnotator.java
@@ -26,6 +26,7 @@ package hudson.plugins.ansicolor;
 
 import hudson.Extension;
 import hudson.MarkupText;
+import hudson.Util;
 import hudson.console.ConsoleAnnotator;
 import hudson.console.ConsoleAnnotatorFactory;
 import hudson.model.Queue;
@@ -35,6 +36,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
 import jenkins.model.Jenkins;
 import org.apache.commons.io.output.CountingOutputStream;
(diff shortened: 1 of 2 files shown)


dwnusbaum:
Trying to trigger a new ci.jenkins.io build now that #169 has been merged, not sure if it will work as-is or if the PR needs to be manually merged with master first.

dwnusbaum:
Oh and it would be nice to go ahead and update CHANGELOG.md in this PR.

jglick:
> update CHANGELOG.md

You are not using Release Drafter?

dwnusbaum:
There are some minor conflicts because #168 was merged, want me to resolve them or do you want to?

jglick:
I will try to fix.

src/main/java/hudson/plugins/ansicolor/ColorConsoleAnnotator.java
```diff
@@ -52,24 +54,33 @@
...
     }
 
     @Override
     public ConsoleAnnotator<Object> annotate(@Nonnull Object context, MarkupText text) {
+        String actualColorMapName;
+        if (colorMapName == null) {
+            actualColorMapName = colorMapNameFor(context);
+            if (actualColorMapName == null) {
+                return this;
+            }
```
dwnusbaum: I'm not sure, but I think the reason that a lot of tests are failing is that `colorMapNameFor(context)` returns null if the build has no `ColorizedAction` (same as previous behavior) or if it does have a `ColorizedAction` but the `colorMapName` is null (different from previous behavior). Previously, in the latter case, the default color map would have been used instead. Maybe it could be fixed by making `colorMapNameFor` return `AnsiColorMap` instead of just the color map name since it can distinguish between the two cases internally.

Issue #172, opened 2020-03-06
color code followed by reset style results in bad output


bandophahita:
The following ansi code combo -- light green with normal style (reset sgr) -- appears to work ok.
`sh "echo -en '\\033[92;22mlightgreen reset sgr \\033[0m\n';"`
![image](https://user-images.githubusercontent.com/4211287/76109169-8bea8f80-5fa1-11ea-9881-38b3f4d8d282.png)

However, when the ansi codes are separated the output appears to break
`sh "echo -en '\\033[92m\\033[22mlightgreen reset sgr \\033[0m\n'"`
![image](https://user-images.githubusercontent.com/4211287/76109275-c3593c00-5fa1-11ea-8ed6-8f3928f3df51.png)

Expected behavior on the second example should match that of the first.

It's worth noting that if any SGR code is set prior to the `\\033[22m` being called, the output works fine.
`sh "echo -en '\\033[92m\\033[1mlightgreen bold \\033[92m\\033[22mlightgreen normal\\033[0m\n';"`
![image](https://user-images.githubusercontent.com/4211287/76109743-b7ba4500-5fa2-11ea-8afb-ce3dcf33487b.png)

The problem appears to be when `22m` is called and no SGR style has been set the output breaks -- but again only when the `22m` is not in a combined ansi code.  
(e.g.`\\033[92m\\033[22m` **bad** vs  `\\033[92;22m` **ok**

Using ansicolor 0.6.2 currently on Jenkins 2.204.1 in a scripted pipeline.

### Test environment

- Jenkins Version: 2.204.1 
- AnsiColor Version: 0.6.2 
- Hostsystem: internal network (no outside access)

### Steps to reproduce the behavior

jenkinsfile scripted example used in regular pipeline. 
```Groovy
node {
    stage('color'){
        ansiColor('xterm'){
            // breaks
            sh "echo -e '\\033[92m\\033[22mlightgreen and separate reset sgr \\033[0m'"
            // works ok
            sh "echo -e '\\033[92;22mlightgreen and combined reset sgr \\033[0m'"
            // works ok
            sh "echo -en '\\033[92m\\033[1mlightgreen bold \\033[92m\\033[22mlightgreen normal\\033[0m'"
        }
    }
}
```

tszmytka:
Thank you for a terrific issue report. I can confirm it is reproducible also in the newest version `0.6.3` so it is definitely a bug.

[...]

PR #173, opened 2020-03-07
172 sgr effect normal intensity

tszmytka:
This is a fix for #172 
Problem was that "SGR Normal intensity" `ESC[22m` was handled correctly **only** if output already contained an html tag `b` opened for "SGR Bold" `ESC[1m` sequence. In other cases it caused no output tags to be generated and the sequence itself was also not commented out.

This PR adds commenting out of any attribute-changing sequence which does not produce output.
Additionally I've added implementation for "SGR Faint" `ESC[2m` similarly to "SGR Bold" `ESC[1m` along with tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 0.6.4 (Next)
 ============
 
+* [#173](https://github.com/jenkinsci/ansicolor-plugin/pull/173): Fix for SGR Normal intensity not handled correctly - [@tszmytka](https://github.com/tszmytka)
 * Your contribution here.
 
 0.6.3 (02/24/2020)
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 0.6.4 (Next)
 ============
 
+* [#173](https://github.com/jenkinsci/ansicolor-plugin/pull/173): Fix for SGR Normal intensity not handled correctly - [@tszmytka](https://github.com/tszmytka)
```
dblock: Needs a period.

tszmytka: > 
> 
> Needs a period.

Added

Issue #156, opened 2019-05-28
isn't there a command to run that can install the plugin


ORESoftware:
In the readme can you list the command at the command line that can install the plugin and enable it?

something like

```bash
$ jenkins-cli install plugin
```

tszmytka:
Added to [README.md](https://github.com/jenkinsci/ansicolor-plugin/blob/master/README.md)

PR #179, opened 2020-05-02
Add instruction for installing via jenkins cli. Improve documentation.

tszmytka:
This PR improves the README by addressing a reasonable (in my opinion) request from #156 and adding status badges concerning:
- `ci.jenkins.io` build
- current plugin version
- amount of installs so far


dblock:
You should fix ci on jenkins before merging this, or we're in broken build forever.

dwnusbaum:
> You should fix ci on jenkins

Good point, I think the failure is caused by changes to `buildPlugin` in https://github.com/jenkins-infra/pipeline-library/pull/92. I think the best fix is to update the parent POM to version 4.1 and update `jenkins.version` to at least 2.176.4 so that dependencies of Jenkins core are resolved using a BOM to avoid enforcer issues for those dependencies.

dblock:
👍

PR #181, opened 2020-05-16
180 eagerness off

### LINKED ISSUE #180: Can't colorize a part of of a Jenkins log.
Please provide precise details of the issue you detected.

### Test environment

- Jenkins Version: 2.222.3
- AnsiColor Version: 0.6.3
- Hostsystem: Linux

### Expected behavior
What did you expect the plugin to do?

Colorize the ansiColor wrapped shell script output ONLY.

### Actual behavior
What did the plugin do instead?

The entire log is colorized.

Create a pipeline job with the below content:

```
pipeline {
    agent { node { label "master" } }
    stages {
        stage('job') {
            steps {
                script {
                    echo "Before Ansi Color Wrap."
                    ansiColor('vga') {
                        sh "ls -la"
                    }
                    echo "After Ansi Color Wrap."
                }
            }
        }
    }
}
```

tszmytka:
This fixes #180

My initial idea of storing the meant-to-be-processed log line numbers would become more complicated than I initially anticipated so I changed it in favor of indicating only the log lines where the `ansicolor-plugin` should either begin or finish processing.

Main bullet points:
- Every time an `ansiColor` step is invoked a `ColorizedAction` step is added to the run context (to indicate starting and stopping)
- At the same time an html-invisible (`div`) marker is injected into the log output to allow `ColorConsoleAnnotator` to recognize when it should do something
- `ColorConsoleAnnotator` inspects each line and if it encounters the marker (matched with a corresponding action) - it either starts/stops its main functionality or simply continues assuming previously configured settings

All tests are passing (including my manual ones) but as there are not that many of them, I would appreciate another opinion.


dblock:
This looks great! I can't find anything wrong with the code ;) Just needs CHANGELOG/README updates and I say go for it. Would bump major version.

tszmytka:
> Just needs CHANGELOG/README updates and I say go for it. Would bump major version.

I keep forgetting about that. Now it is also there. 
I've also added a README example for the feature and moved the build wrapper pipeline usage before the old-fashioned UI defined job as this probably the preferred way of using it in `Jenkinsfile` times.

PR #184, opened 2020-05-24
Use GitHub README as plugins.jenkins.io docs

MarkEWaite:
The GitHub README is more detailed than the Jenkins wiki page and provides a better experience for the reader.

See the [blog post](https://www.jenkins.io/blog/2019/10/21/plugin-docs-on-github/) and the [documentation page](https://www.jenkins.io/doc/developer/publishing/documentation/#documenting-plugins) for more details.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -15,7 +15,7 @@
 
   <name>AnsiColor</name>
   <description>Adds ANSI coloring to the Console Output</description>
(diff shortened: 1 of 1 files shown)


tszmytka:
Thank you for this.
I was actually thinking that it would be nice if the wiki got automatically updated after changes to the `README` and wasn't aware that this was already possible.

Jenkins build failed but it seems incidental.

PR #190, opened 2020-06-05
Leaky formatting

tszmytka:
This is a fix for [JENKINS-61598](https://issues.jenkins-ci.org/browse/JENKINS-61598)

It probably got introduced as a side effect of #137

### DIFF
--- a/src/test/java/hudson/plugins/ansicolor/action/ColorizedActionTest.java
+++ b/src/test/java/hudson/plugins/ansicolor/action/ColorizedActionTest.java
@@ -60,9 +60,18 @@ public class ColorizedActionTest {
     }
 
     @Test
-    public void willReturnDefaultIfLogAnnotationPointsToNonexistingAction() {
+    public void willReturnDefaultIfLogAnnotationPointsToNonexistentAction() {
         final MarkupText markupText = new MarkupText("Log line");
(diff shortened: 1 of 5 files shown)


src/test/java/hudson/plugins/ansicolor/action/ColorizedActionTest.java
```diff
@@ -60,9 +60,18 @@ public void willReturnDefaultIfLogDoesntContainAnnotation() {
...
         final MarkupText markupText = new MarkupText("Log line");
         markupText.addMarkup(0, TAG_ACTION_BEGIN + "\"identifier_not_in_actions\"" + TAG_ACTION_END);
         assertEquals(CONTINUE, ColorizedAction.parseAction(markupText, buildRun));
     }
+
+    @Test
+    public void willReturnCommandIgnoreOnPipelineInternalLine() {
+        final MarkupText markupText = new MarkupText("Some internal line");
+        markupText.addMarkup(0, "<span class=\"pipeline-new-node\">");
+        System.out.println(markupText.toString(false));
```
dblock, line 73: Debugging? Remove.

tszmytka: Oh the shame ;-)
Believe it or not I **never** do this.
Removed.

PR #191, opened 2020-06-05
Validate configuration form field values before storing them.

tszmytka:
This introduces configuration form input field validation for `ansicolor` settings available under `/configure`. Most of the fields had no rules defined.
Specifically the PR adds the following:
- each custom color map name: 1-256 chars (till now it could be any length >0)
- each custom color literal: 1-64 chars (till now there was no validation)
- global color map name must match one of the color maps (till now it could be any string)
- setting global color map name to an empty string removes it entirely from settings (till now the empty string was saved as-is) - the same state as in a clean Jenkins installation

### DIFF
--- a/src/main/java/hudson/plugins/ansicolor/AnsiColorBuildWrapper.java
+++ b/src/main/java/hudson/plugins/ansicolor/AnsiColorBuildWrapper.java
@@ -49,7 +49,7 @@ import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.function.Supplier;
+import java.util.function.Function;
 
 /**
  * Build wrapper that decorates the build's logger to filter output with {@link AnsiHtmlOutputStream}.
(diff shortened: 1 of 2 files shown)


src/main/java/hudson/plugins/ansicolor/AnsiColorBuildWrapper.java
```diff
@@ -117,19 +115,139 @@ private void addAll(AnsiColorMap[] maps, Map<String, AnsiColorMap> to) {
...
-            return (value.trim().length() == 0) ? FormValidation.error("Name cannot be empty.") : FormValidation.ok();
+            return validateColorMapName(value);
+        }
+
+        private FormValidation validateColorMapName(String name) {
+            final int nameLength = name.trim().length();
+            return (nameLength < 1 || nameLength > 256) ? FormValidation.error("Color map name length must be between 1 and 256 chars.") : FormValidation.ok();
+        }
+
+        public FormValidation doCheckBlack(@QueryParameter String value) {
```
dblock, line 184: How do these unused ones get invoked?

src/main/java/hudson/plugins/ansicolor/AnsiColorBuildWrapper.java
```diff
@@ -117,19 +115,139 @@ private void addAll(AnsiColorMap[] maps, Map<String, AnsiColorMap> to) {
         @Override
         public boolean configure(final StaplerRequest req, final JSONObject formData) throws FormException {
             try {
-                setColorMaps(req.bindJSONToList(AnsiColorMap.class,
-                        req.getSubmittedForm().get("colorMap")).toArray(new AnsiColorMap[1]));
-                setGlobalColorMapName(req.getSubmittedForm().getString("globalColorMapName"));
+                final List<AnsiColorMap> colorMaps = req.bindJSONToList(AnsiColorMap.class, req.getSubmittedForm().get("colorMap"));
+                for (AnsiColorMap colorMap : colorMaps) {
+                    validateFieldName(colorMap.getName());
+                    validateFieldColorLiteral(colorMap.getBlack(), "black");
```
dblock, line 121: I would swap value and name, feels weird that it's value, then name, but no big deal.

Maybe there's a way to loop over these?

src/main/java/hudson/plugins/ansicolor/AnsiColorBuildWrapper.java
tszmytka: That's Jenkins magic. All the `doCheck*` methods get called automatically on form data submission for every field with a matching method.

src/main/java/hudson/plugins/ansicolor/AnsiColorBuildWrapper.java
tszmytka: > I would swap value and name

I started off with the value alone and only later added the field name. Maybe it does make more sense to have the name on the first spot.

> Maybe there's a way to loop over these?

There's not much that can be done here I'm afraid. Maybe putting the field names and method references in a map and looping through that. Do you think this approach looks better?

src/main/java/hudson/plugins/ansicolor/AnsiColorBuildWrapper.java
```diff
@@ -118,22 +120,26 @@ public boolean configure(final StaplerRequest req, final JSONObject formData) th
...
-                    validateFieldColorLiteral(colorMap.getYellowB(), "yellowB");
-                    validateFieldColorLiteral(colorMap.getBlue(), "blue");
-                    validateFieldColorLiteral(colorMap.getBlueB(), "blueB");
-                    validateFieldColorLiteral(colorMap.getMagenta(), "magenta");
-                    validateFieldColorLiteral(colorMap.getMagentaB(), "magentaB");
-                    validateFieldColorLiteral(colorMap.getCyan(), "cyan");
-                    validateFieldColorLiteral(colorMap.getCyanB(), "cyanB");
-                    validateFieldColorLiteral(colorMap.getWhite(), "white");
-                    validateFieldColorLiteral(colorMap.getWhiteB(), "whiteB");
+                    final HashMap<String, Supplier<String>> validatedFields = new HashMap<>();
```
dblock, line 123: I think this belongs in some other static method that returns the map of validated fields.

src/main/java/hudson/plugins/ansicolor/AnsiColorBuildWrapper.java
dblock: I like what you did to this plus I would extract that map in a method.

dblock:
Looking forward to some tests and this is good to go.

src/main/java/hudson/plugins/ansicolor/AnsiColorBuildWrapper.java
tszmytka: Moved it to a static Map.

src/main/java/hudson/plugins/ansicolor/AnsiColorBuildWrapper.java
tszmytka: Did that and covered with tests.

dblock:
Merged.

dblock:
This should probable have had a changelog entry. Feel free to commit one.

PR #196, opened 2020-07-12
193 interaction timestamper

tszmytka:
This adjustment addresses the problem of `timestamper` plugin decorating the global `logger` with functionality that writes anything to the log **only** if the line ends with a `LF` character, otherwise it gets lost.
The `ansicolor` annotates log lines where its own logic should start/stop with `ConsoleNote`s which get serialized without any special line ending. If those lines are not available in the log upon rendering the whole coloring is not enabled and any special chars end up in the output it their raw form.


dblock:
Looks hacky :) Feel free to merge on green.

tszmytka:
> 
> 
> Looks hacky :) Feel free to merge on green.

It is, I know, but there really is no clean way to work around this problem without adjustments in `timestamper`.

Jenkins must have had hiccup. Additional push and all builds are green.

PR #202, opened 2020-08-29
Render logs correctly for long running jobs which generate lots of logs

### LINKED ISSUE #200: Raw escape sequences in output while refreshing a long running build
There's still a situation where it's not working correctly so I made up a somewhat contrived pipeline script.

```groovy
pipeline {
  agent any

  options {
    ansiColor('xterm')
  }

  stages {
    stage('Run') {
      steps {
          echo '\033[32mcolored output\033[0m'
sh label: 'colors', script: '''echo start
for c in {30..37}; 
do
  printf "Time for $c\\n"
  sleep 3
  echo -e "\\e[31mHello World\\e[0m"
  echo -e "\\e[${c}m Loop $c \\e[0m"
  printf '\\033[32mcolored output\\033[0m\\n'
  printf "\\033[${c}mcolored output\\033[0m\\n" 
  for i in {000..199}
  do 
    printf "\\033[${c}mI\\\'m filler $i taking up a bunch of log space ####################################################################################\\033[0m\\n" 
  done
  echo next
done'''
    
    
      }
    }
  }
}
```

The problem for me now is when Jenkins truncates the output when it gets too long. 
After starting that pipeline mash refresh on the console output view every ~3 seconds. It's looping through colors 30 to 37 and at about 33 (refresh when it's on green) refreshing loses the color. I suspect it's hitting a case were an escape sequence is getting split by the console getting truncated and whatever is doing the parsing decides not to try past the invalid input but just a guess.
___
Peek 2020-08-05 10-40 ansiColor
Example where I hit refresh on the green step (there colors stop working) and then after it finishes (color works again).

Versions:
```
Jenkins 2.235.1
ansicolor 0.7.2
```

_Originally posted by @chder in https://github.com/jenkinsci/ansicolor-plugin/issues/193#issuecomment-669268756_

tszmytka:
Fixes #200 
All is pretty much described in the issue.
The PR adds a way for checking the currently active color map name in the currently running job run.

### DIFF
--- a/src/main/java/hudson/plugins/ansicolor/action/ColorizedAction.java
+++ b/src/main/java/hudson/plugins/ansicolor/action/ColorizedAction.java
@@ -49,16 +49,13 @@ public class ColorizedAction extends InvisibleAction {
     private final Command command;
 
     public enum Command {
-        /**
-         *
-         */
         START,
         STOP,
         CONTINUE,
         IGNORE,
 
         /**
-         * Currently running execution
+         * The running job's current action containing current color map name
          */
         CURRENT
     }
--- a/src/main/java/hudson/plugins/ansicolor/action/ColorizedAction.java
+++ b/src/main/java/hudson/plugins/ansicolor/action/ColorizedAction.java
@@ -31,6 +31,7 @@ import hudson.plugins.ansicolor.AnsiColorMap;
 
 import java.util.Optional;
 import java.util.UUID;
+import java.util.logging.Logger;
 
 import static hudson.plugins.ansicolor.action.ActionNote.TAG_ACTION_BEGIN;
 
@@ -38,6 +39,7 @@ import static hudson.plugins.ansicolor.action.ActionNote.TAG_ACTION_BEGIN;
  * Action for issuing commands to ColorConsoleAnnotator
  */
 public class ColorizedAction extends InvisibleAction {
+    private static final Logger LOGGER = Logger.getLogger(ColorizedAction.class.getName());
     private static final String TAG_PIPELINE_INTERNAL = "<span class=\"pipeline-new-node\"";
     static final ColorizedAction CONTINUE = new ColorizedAction("", Command.CONTINUE);
     static final ColorizedAction IGNORE = new ColorizedAction("", Command.IGNORE);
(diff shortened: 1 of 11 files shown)


tszmytka:
The new test is completing fine on my localhost but failing on both CI hosts.
Let me figure out what is going on.

src/main/java/hudson/plugins/ansicolor/action/ColorizedAction.java
```diff
@@ -29,14 +29,17 @@
...
 import java.util.UUID;
+import java.util.logging.Logger;
 
 import static hudson.plugins.ansicolor.action.ActionNote.TAG_ACTION_BEGIN;
 
 /**
  * Action for issuing commands to ColorConsoleAnnotator
  */
 public class ColorizedAction extends InvisibleAction {
+    private static final Logger LOGGER = Logger.getLogger(ColorizedAction.class.getName());
```
dblock, line 42: Looks like code uses `LOG` and `LOGGER` in other places, for the future maybe make them consistent. 

I prefer `LOGGER`.

src/main/java/hudson/plugins/ansicolor/action/ColorizedAction.java
```diff
@@ -87,7 +95,20 @@ public static ColorizedAction parseAction(MarkupText text, Run<?, ?> run) {
             final String id = line.substring(from, to);
             return run.getActions(ColorizedAction.class).stream().filter(a -> id.equals(a.getId())).findAny().orElse(CONTINUE);
         }
-        return line.contains(TAG_PIPELINE_INTERNAL) ? IGNORE : CONTINUE;
+        if (line.contains(TAG_PIPELINE_INTERNAL)) {
+            return IGNORE;
+        }
+        final boolean isBuilding = run.isBuilding();
+        LOGGER.fine("Run is building: " + isBuilding);
```
dblock, line 102: I've seen `.finest` elsewhere, same as the comment above, make it consistent. Potentially remove logging where it's not useful anymore and was there fore debugging.

dblock:
Add an entry to https://github.com/jenkinsci/ansicolor-plugin/blob/master/CHANGELOG.md and this is good to squash-merge.

src/main/java/hudson/plugins/ansicolor/action/ColorizedAction.java
tszmytka: > I prefer `LOGGER`.

Agreed

src/main/java/hudson/plugins/ansicolor/action/ColorizedAction.java
tszmytka: I used it because only `fine` or more severe messages are show by default in failed test output on `https://ci.jenkins.io/`. I will gladly get rid of this entirely but I think it's worth waiting a bit and allowing the whole project to be built a couple of times more - if the same test starts failing again the output will be there from the beginning.

PR #213, opened 2020-12-29
Remove tabular markup on new Jenkins core

timja:
Following https://www.jenkins.io/doc/developer/views/table-to-div-migration/

See https://www.jenkins.io/blog/2020/11/10/major-changes-in-weekly-releases/

I'm not sure if this is causing an issue but I'm working through all the plugins that appear with tabular markup in https://issues.jenkins.io/browse/JENKINS-64341

cc @tszmytka

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>4.7</version>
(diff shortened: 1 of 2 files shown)


tszmytka (approved):
Changes look ok.
Manually tested with `2.204.6` (got tables) and `2.273` (got divs). No problems with editing/saving form data both times.

timja:
@tszmytka could you release it please? ❤️

PR #219, opened 2021-02-06
218 term

### LINKED ISSUE #218: env.TERM/TERM not set if global config is e.g. "xterm"
i was running a declarative Pipeline on a Jenkins master with
a) the global configuration for ansicolor-plugin at '' value
b) the same with 'xterm' value
c) counter checking with some job local setting of ansiColor('xterm')

for case a) i found all to work as expected (env.TERM was null, sh "echo $TERM" returned 'dumb'; ansi color sequences were left as it is).
for case b) i found something not right: env.TERM was null, sh "echo $TERM" returned 'dumb'; but ansi color sequences were honored.
for case c) i round all to be as expected

### Test environment
```
- Jenkins Version: 2.235.5
- AnsiColor Version: 0.7.3
- Operating System running Jenkins: Linux Ubuntu 16.04
```

### Expected behavior
for case b) I expected the various levels of environment variable storage to be set on a global level

### Actual behavior
for case b) the environment variables were left at the defaults.

see inital paragraph

### Issue checklist
- [x] Jenkins version
- [x] AnsiColor version
- [x] Operating System name and version
- [x] Expected behavior
- [x] Actual behavior
- [x] Steps to reproduce the behavior

tszmytka:
Fixes #218

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 0.7.6 (Next)
 ============
 
+* [#219](https://github.com/jenkinsci/ansicolor-plugin/pull/219): Set env var TERM also if a default color map is used - [@tszmytka](https://github.com/tszmytka).
 * Your contribution here.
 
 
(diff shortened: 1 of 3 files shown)


tszmytka:
I've squashed the "Temp configure to run on basic infra" commits - the build kept on failing on `ci.jenkins.io` and I wasn't sure what the problem was. Turns out there wasn't any 😕 - it got back to green after I bumped it the nth+1 time.

PR #220, opened 2021-03-09
Fixing jenkinsci#216: partial logs not showing colors

### LINKED ISSUE #216: partial logs not showing colors
follow to #212 
test
- Jenkins Version:2.263.1
- AnsiColor Version: compiled the branch of [tszmytka:212-partial-log-output ](https://github.com/jenkinsci/ansicolor-plugin/pull/214)
- Operating System running Jenkins: Ubuntu 14.04.6 LTS

repreduces:
free style job
```
#!/bin/bash
set -x
for i in {1..5000}; do  
echo -e '[0;37;42m[10][0m [32m  my letters'
done
```
image

image

roehrijn:
I think this fixes #216. At least it does for my setup. But now the test hudson.plugins.ansicolor.ColorConsoleAnnotatorTest#testNoGlobalPipelineColorMap is failing and I do not fully understand about the mechanism behind. 

Problem seems to be a mixture of a missing pipeline step log line which could define the color map name in combination with partial / progressive output. Whenever there is no log line originated from a pipeline step rendered before the first ANSI escape sequence, the color map stays null and the log lines are not converted to (colorful) HTML. This often happens when a lot of colorful logs are written between to pipeline log lines or no pipeline log line at all. 

I thought it could be worth providing this PR even though the mentioned test is failing. I hope this can help to start a discussion and I can bring it towards a useful and sustainable fix.

### DIFF
--- a/src/main/java/hudson/plugins/ansicolor/ColorConsoleAnnotator.java
+++ b/src/main/java/hudson/plugins/ansicolor/ColorConsoleAnnotator.java
@@ -89,6 +89,7 @@ final class ColorConsoleAnnotator extends ConsoleAnnotator<Object> {
         switch (colorizedAction.getCommand()) {
             case START:
             case CURRENT:
+            case CONTINUE:
                 colorMapName = colorizedAction.getColorMapName();
                 break;
             case STOP:


tszmytka:
Nice try but this really doesn't fix anything. Much too soon for a PR.
Your change enables the plugin's functionality for most of the output lines no matter if ansicolor is enabled or not which is a pretty important distinction. 
Apart from the failing tests it would definitely cause a few additional issues from people using it to handle only part of their logs (or only specific lines).
As I wrote in #216 - what would be really helpful right now is a script or a pipeline reproducing the issue.

roehrijn:
I think it is never too soon for a PR. I never expected this to be merged because I new that at least the tests are failing. As I just stated, I wanted to pinpoint the are where the problem is arising.

PR #227, opened 2021-05-15
Trigger tailed log output colorization also after Jenkins restart

tszmytka:
Currently the functionality allowing tailed log output to be colorized is executed at the very end of a build - after the log file has already been closed (`onFinalized`). This permits accurate calculation of the point where tailed output will begin and correct `ColorizedAction.Command.START` action injection. Upon rendering, the new action gets added to the output but since the log file has already been closed the new content is not persisted. As a result tailed log output looks fine until a Jenkins instance gets rebooted. After the restart it gets shown without any ansi code interpreted (full output is fine before and after reboot).

The PR fixes this by moving the execution a bit earlier - to `onCompleted`. This point in the lifecycle of a `WorkflowRun` happens before log file has been closed so any log amendments get properly persisted and tailed log output gets correctly rendered before and after Jenkins restart. One drawback is that one final line gets added to the log *after* `onCompleted` so tailed output begin calculation needs to take that into consideration.
This change _might_ also work for #216 but I can't be 100% sure as I haven't been able to reproduce that issue.

- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your master branch!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 * [#219](https://github.com/jenkinsci/ansicolor-plugin/pull/219): Set env var TERM also if a default color map is used - [@tszmytka](https://github.com/tszmytka).
 * [#226](https://github.com/jenkinsci/ansicolor-plugin/pull/226): Work around withMaven step not passing on log line metadata - [@tszmytka](https://github.com/tszmytka).
 * [#225](https://github.com/jenkinsci/ansicolor-plugin/pull/225): Work around logstash-plugin hiding some log lines from ansicolor - [@tszmytka](https://github.com/tszmytka).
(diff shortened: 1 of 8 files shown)


tszmytka:
When this gets merged I think it will be high time to release a new version and with that I would like to propose bumping the version to a full `1.0.0` to indicate the project's well-above-infancy status.

Function-wise `ansicolor-plugin` has been stable for a long period of time covering at least most of the use-cases. Personally I think this bump could have been easily made years ago.

Keeping the version on a pre-1.0 level doesn't reflect the real maturity of the project and may very well discourage potential users from installing it (Why is the version < 1.0? Is it in a beta state? Maybe it is in experimental shape and not really recommended for real-world use yet?).

PR #238, opened 2022-01-07
EOL JSR 305

basil:
Compatible with [plugin POM 4.32](https://github.com/jenkinsci/plugin-pom/releases/tag/plugin-4.32).


dblock:
Does this change anything in terms of compatibility for users? Maybe add a line to https://github.com/jenkinsci/ansicolor-plugin/blob/master/CHANGELOG.md either way. Thanks!

basil:
> Does this change anything in terms of compatibility for users?

No, it does not.

> Maybe add a line to https://github.com/jenkinsci/ansicolor-plugin/blob/master/CHANGELOG.md either way.

There is no user-visible change.

Issue #249, opened 2022-08-03
Documentation does not describe what magic strings 'xterm' and 'css' does


tarjei:
### Describe your use-case which is not covered by existing documentation.

Hi, I tried to search for it but could not find any place where the difference between 

`ansiColor('xterm')` and `ansiColor('css')` was described or elaborated. This would be helpfull.

dblock:
These are color maps, https://github.com/jenkinsci/ansicolor-plugin/blob/master/src/main/java/hudson/plugins/ansicolor/AnsiColorMap.java#L45

Please do contribute documentation about how this works!

PR #250, opened 2022-08-05
Describe what the argument to ansiColor() does.

tarjei:
- [x] Link to relevant issues in GitHub or Jira - this is related to #249 
- [ ] Link to relevant pull requests, esp. upstream and downstream changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ ansiColor('xterm') {
 
 ## Colormaps
 
-The argument to the ansiColor function can be either `term`, `vga`, `gnome-terminal` or `css`. This defines which colors Jenkins should show for some of the escape codes. You can find the exact colors used here: https://github.com/jenkinsci/ansicolor-plugin/blob/,aster/src/main/java/hudson/plugins/ansicolor/AnsiColorMap.java#L45.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -73,6 +73,10 @@ ansiColor('xterm') {
 }
 ```
 
+## Colormaps
+
+The argument to the ansiColor function can be either `term`, `vga`, `gnome-terminal` or `css`. This defines which colors Jenkins should show for some of the escape codes. You can find the exact colors used here: https://github.com/jenkinsci/ansicolor-plugin/blob/,aster/src/main/java/hudson/plugins/ansicolor/AnsiColorMap.java#L45.
```
dblock, line 78: Let's make this a markdown link? `[here](link)`, and fix the `,aster` typo? You don't need to include github.... here either, these links are relative, so `src/main/java/hudson/plugins/ansicolor/AnsiColorMap.java` will do. I would probably exclude `#L45` because that file might change in the future, but up to you.

dblock (changes requested):
See comment, the typo is a must fix.

README.md
dblock: Are these color maps some kind of standard? Maybe we can research that and include a link?

eightnoneone: Avoid linking the word "here". Hypertext links should add information to a word in context.

> This defines which set of colors Jenkins should show for some of the escape codes. The exact color values used in each set are defined in [AnsiColorMap.java](src/main/java/hudson/plugins/ansicolor/AnsiColorMap.java).

tarjei:
@dblock good comments! Thanks for taking the time :) 

I've fixed the typo + changed the wording for the link.

dblock:
Thanks for finishing this! Merged.

PR #253, opened 2023-02-24
Refresh dependencies and enable dependabot

jonesbusy:
Hi,

Build against modern Jenkins (drop Java 8 support)

Regards,

- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [x] Ensure that the pull request title represents the desired changelog entry
- [x] Please describe what you did

PR #254, opened 2023-02-25
Refresh plugin 2023

dblock:
Same as #253 but by a maintainer.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,8 @@
+version: 2
+updates:
+- package-ecosystem: maven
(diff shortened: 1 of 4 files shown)


dblock:
@jonesbusy these look like legit failures

jonesbusy:
> @jonesbusy these look like legit failures

At first look is weird. Tests are passing on Linux but not Windows

dblock:
If anyone wants to continue this, https://github.com/jenkinsci/ansicolor-plugin/pull/258 might get you further.

PR #306, opened 2024-11-11
Use Jenkins theme CSS color variables for inverse text

jdknight:
The original support for negative/inverse text uses a combination off `currentColor` and a bright-level-7 value. When using a default Jenkins theme, these color definitions work as expected. However, clients using the dark theme may not be able to read the text since both the text and background are a bright color.

Modern Jenkins themes provide CSS color variables. Instead of trying to rely on `currentColor`, use the default text color (`var(--text-color)`) and the default background color (`var(--background)`). This should allow negative/inverse text to render appropriately no matter what theme mode is used.

### Testing done

With unit test changes that seemed to be right (please check), perform a `mvn test` to confirm expected results.

Started up a fresh Jenkins container instance and setup a pipeline job with the following:

```
node {
    stage('Build') {
        ansiColor('xterm') {
            echo 'This is a test.'
            echo '\033[7mThis is a test.\033[0m'
            echo 'This is a test.'
        }
    }
}
```

Before the changes, installed the ansicolor plugin from the online and observed the issue:

image

Then uninstalled the plugin, restarted and manually installed a newly built `ansicolor.hpi` plugin. Navigated back to the job results and observed that the inversed text could be read:

image

### Submitter checklist
- [ ] Link to relevant issues in GitHub or Jira
- [ ] Link to relevant pull requests, esp. upstream and downstream changes
- [ ] Ensure you have provided tests - that demonstrates feature works or fixes the issue

### DIFF
--- a/src/main/java/hudson/plugins/ansicolor/AnsiHtmlOutputStream.java
+++ b/src/main/java/hudson/plugins/ansicolor/AnsiHtmlOutputStream.java
@@ -313,27 +313,8 @@ public class AnsiHtmlOutputStream extends AnsiOutputStream {
         Integer defaultFgIndex = colorMap.getDefaultForeground();
         if (defaultFgIndex != null) color = colorMap.getNormal(defaultFgIndex);
         if (color == null) {
-            // with no default foreground set, we need to guess about (currently happened in xterm and css themes)
-            // possible approaches are:
-            // • hardcoded black "#000000"
-            // • colorMap.getNormal(0)
-            // • "currentColor" → see also: http://stackoverflow.com/a/42586457/2880699
-            //   but note, that this approach only works, if all <span style="color: …"> are currently closed
-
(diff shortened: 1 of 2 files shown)


dblock:
Very nice! Add a line to https://github.com/jenkinsci/ansicolor-plugin/blob/master/CHANGELOG.md? 

Let's increment the version to 1.1 as part of this PR as well, it's a big change.

jdknight:
> Add a line to https://github.com/jenkinsci/ansicolor-plugin/blob/master/CHANGELOG.md?

I do not mind adding some information there, but it seemed that as of v1.0.3, change log entries are [only being populated inside GitHub releases](https://github.com/jenkinsci/ansicolor-plugin/commit/42f4c774b6462c3f044823739c83e6eb74bf0688).

dblock:
> Let's increment the version to 1.1 as part of this PR as well, it's a big change.

@jdknight want to do this one before I merge? and sorry for the delay

PR #353, opened 2025-07-29
Migrate from Commons Lang 2 to native Java Platform functionality

basil:
No need to use a third-party library for this when the Java Platform provides this functionality natively.


basil:
This extremely minor internal change doesn't seem to merit a changelog entry. Other parts will still use Commons Lang 2 until someone decides to remove it from core.

PR #378, opened 2026-03-29
Add usage section to README

jahanvipandey22:
Added a basic usage section to the README to help beginners understand how to use the AnsiColor plugin.
This improves documentation clarity for new users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ and has [a page](https://wiki.jenkins-ci.org/display/JENKINS/AnsiColor+Plugin) o
 ```
 java -jar jenkins-cli.jar install-plugin ansicolor
 ```
-## Use
+## Enable and Usage
 
 After installing the AnsiColor plugin:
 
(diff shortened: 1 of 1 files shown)


dblock (changes requested):
Thanks! Let's move it up after installation and call it "Use"?

jahanvipandey22:
Hi @dblock, I have updated and repositioned the section as suggested. Kindly review.

dblock (commented):
Sorry to be annoying, but we have an Enable section. Is this different? Should this be a section under Enable with a new name?

jahanvipandey22:
Thanks for the suggestion!

You're right — the previous "Usage" section overlapped with the existing "Enable" section. I've now merged them into a single "Enable and Usage" section and reorganized the content accordingly.

Please let me know if this looks good now!

dblock:
We actually have it here: https://github.com/jenkinsci/ansicolor-plugin/blob/master/README.md#in-a-traditional-job

Feel free to reorganize the README if it's not clear.

